PR #113, opened 2025-11-14
add draft for volcanoplot wrapper function

vbrennsteiner:
[UPDATE] --> new logic using `plot_layers` list instead of `layer_dict`

By popular and LLM demand, there should be a volcanoplot wrapper function in alphatools. As such, it should abstract the following functionalities:

- scatterplot
- adding lines to the scatterplot
- adding anchored labels to the scatterplot
- adding axis labels (this is debatable, maybe we could keep this external)
- adding a legend
- setting plot x/y limits

A particular consideration for volcanoplots is the differential coloring of various points of interest,
e.g. significantly regulated, manually selected or belonging to a certain group of interest. The current
solution uses `plot_layers` to define the hierarchy of visualization layers. `plot_layers` is a list of
tuples, where each tuple contains

layer_column, color_key, layer_val

- a `layer_column` from `data`
- a `color_key` to retrieve a color from the `color_dict`
- a `layer_val` to match in the `layer_column`, either a single value or a list of values

For example, given this plot_layers definition:

```
pois = ["P10291", "P10292", "P10293", "P10294", "P10295"]
plot_layers = [

    # look for all points in the "id" column that are in the pois list and color them according to the "POI_hypothesis" color in the color_dict
    ("id", "POI_hypothesis", pois), 

    # look for all points that have "POI" in the "label" column and color them according to the "POI" color in the color_dict
    ("label", "POI", "POI"), 

    # similar for regulation status
    ("diff_exp_status", "upregulated", "upregulated"),
    ("diff_exp_status", "downregulated", "downregulated"),
    ("diff_exp_status", "unchanged", "unchanged"),
]
```

Importantly, order matters: these plot_layers ensure that the 'pois' are on top, then the points labeled as 'POI', then the upregulated, downregulated and unchanged points in the background.

Combined with this layer_dict, we specify that we want to add labels to POI points and POI_hypothesis points.

```
label_layers = [
    "POI", # label all points where the label is "POI"
    "POI_hypothesis", # label all points in the pois list
]
```

in this configuration, every point is plotted exactly once in one layer, meaning there are no overplotting artefacts that sometimes show up in PNG/SVG renderings when multiple versions of a point with different colors are positioned on top of each other.

Means "At the first layer, color those points where the `Genes` column's values are in genes_of_interest in black. At second layer, color the points based on their status in the `diff_exp_status` column.". This implementation solves overplotting problems, since users specify the hierarchy in the order of the layer_dict (i.e. selected genes will always be plotted on top, differential expression status below that, etc. for further layers). Additionally, every point is plotted exactly once in one layer, meaning there are no overplotting artefacts that sometimes show up in PNG/SVG renderings when multiple versions of a point with different colors are positioned on top of each other.

### Example:

```
# Select some points we want to highlight, e.g. if these had some hypothesis attached
pois = ["P10291", "P10292", "P10293", "P10294", "P10295"]

plot_layers = [
    ("id", "POI_hypothesis", pois),  # in the first layer, plot all points in pois list
    ("label", "POI", "POI"),  # in the second layer, plot all points where the label is "POI"
    ("diff_exp_status", "upregulated", "upregulated"),  # at the third layer
    ("diff_exp_status", "downregulated", "downregulated"),  # in the fourth layer
    ("diff_exp_status", "unchanged", "unchanged"),  # in the fifth layer
]

color_dict = {
    "upregulated": BaseColors.get("orange"),  # the specific point "P10001" should be colored orange
    "downregulated": BaseColors.get("blue"),  # the specific point "P10002" should be colored blue
    "unchanged": BaseColors.get("grey"),  # points from the list ["P10002", "P10003", "P10004"] should be colored grey
    "POI": "black",  # points where the label is "POI" should be colored black
    "POI_hypothesis": BaseColors.get("purple", lighten=0.7),  # points in the pois list should be colored purple
}

# Add labels to selected points
label_layers = [
    "POI",  # label all points where the label is "POI"
    "POI_hypothesis",  # label all points in the pois list
]


Plots.volcano(
    data=vp_data,
    x_column="log2fc",
    y_column="neg_log10pval",
    color_dict=color_dict,
    plot_layers=plot_layers,
    label_layers=label_layers,
    x_label_anchors=[-3.5, 3.5],
    y_padding_factor=1.7,
    y_display_start=0.75,
    xlims=(-6, 6),
    figure_kwargs={"figsize": (5, 3)},
)
```


src/alphapepttools/pl/plots.py
```diff
@@ -1442,3 +1442,133 @@ def plot_pca_loadings_2d(
...
+) -> None:
+    """Visualize 🅱️olcano plot"""
+    scatter_kwargs = scatter_kwargs or {}
+    label_kwargs = label_kwargs or {}
+    figure_kwargs = figure_kwargs or {"figsize": (6, 4)}
+
+    def _tolist(
+        obj: str | list,
+    ) -> list:
+        return obj if isinstance(obj, list) else [obj]
```
mschwoer: saw this yesterday elsewhere around .. -> public helper method?

src/alphapepttools/pl/plots.py
```diff
@@ -1442,3 +1442,133 @@ def plot_pca_loadings_2d(
...
+
+    X_COLS = ["log2fc"]
```
mschwoer: why ALL_UPPER?

src/alphapepttools/pl/plots.py
```diff
@@ -1442,3 +1442,133 @@ def plot_pca_loadings_2d(
...
+    if ax is None:
+        fig, ax = create_figure(1, 1, **figure_kwargs)
+        ax = ax.next()
+
+    ### Add visualization layers ###
+    glob_spent_idxs = []
+    glob_layer_idxs = []
+    for layer_column, layer_values in layer_dict.items():
+        for value in _tolist(layer_values):
+            layer_colorname = layer_column if isinstance(value, list) else value
```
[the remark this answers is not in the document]
mschwoer: I think I need some explanation on that `layer_dict` idea

src/alphapepttools/pl/plots.py
```diff
@@ -1442,3 +1442,133 @@ def plot_pca_loadings_2d(
         # set axis labels
         space_suffix = " (samples)" if dim_space == "obs" else " (features)"
         label_axes(ax, xlabel=f"PC{pc_x}{space_suffix}", ylabel=f"PC{pc_y}{space_suffix}")
+
+
+def diff_exp_volcano(  # NOQA: C901
```
lucas-diedrich: Just `alphatools.pl.volcano`?

src/alphapepttools/pl/plots.py
```diff
@@ -1442,3 +1442,133 @@ def plot_pca_loadings_2d(
...
+        return obj if isinstance(obj, list) else [obj]
+
+    X_COLS = ["log2fc"]
+
+    if x_column not in X_COLS:
+        raise ValueError(f"x_column must be one of {X_COLS}")
+
+    # Determine sensible limits
+    valid_x_values = ttest_df.loc[((~ttest_df[x_column].isna()) & (~ttest_df[y_column].isna())), x_column]
+    abs_max_x = max(abs(valid_x_values.min()), abs(valid_x_values.max()))
```
lucas-diedrich: Why not 

```python
valid_x_values.abs().max()
```

src/alphapepttools/pl/plots.py
```diff
@@ -1442,3 +1442,133 @@ def plot_pca_loadings_2d(
...
+
+    fig = None
+    if ax is None:
+        fig, ax = create_figure(1, 1, **figure_kwargs)
+        ax = ax.next()
+
+    ### Add visualization layers ###
+    glob_spent_idxs = []
+    glob_layer_idxs = []
+    for layer_column, layer_values in layer_dict.items():
```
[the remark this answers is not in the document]
lucas-diedrich: Note that just because we remove custom plotting functionalities, we can still very much define the style of the plots (with the pre-specified palettes, and other settings), but reduce the complexity of basic plotting functionality

src/alphapepttools/pl/plots.py
vbrennsteiner: whoops, yes that's a little more verbose than it needs to be 😆

src/alphapepttools/pl/plots.py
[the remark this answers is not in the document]
vbrennsteiner: Interesting! If this is something we can use ready made then it would be great, as long as we have a flexible solution to set everything as we need - this would finally add seaborn to our dependencies after all, but let's look into it - one thing I'm not understanding yet is how we would then color by different columns and their respective categories, as in 

```
      layer_dict={
          id_column: [top_N],
          "kinase_status": ["kinase"],
          "diff_exp_status": ["upregulated", "downregulated", "unchanged"],
      },
```

I understand that we could call seaborn.scatterplot() repeatedly with the respective entries, but how would we ensure that everything is plotted exactly once without the plot layer logic?

PR #164, opened 2026-01-27
Extend pl module docstrings

vbrennsteiner:
### Reason for this PR:

While many docstrings are already adapted for automated documentation, many function documentations are still spotty and therefore harder to pares/use for humans and LLMs. This is the first of following PRs aiming to increase docstring quality by updating the description and parameters and adding code examples based on the tutorial notebooks for each function.

### Breaking change:

the interface of `label_plot` changes from using np.ndarray / pd.Series to using pd.DataFrame or ad.AnnData with column names as input. This makes its interface consistent with our remaining dataframe/anndata based functions.


lucas-diedrich (commented):
As discussed I would split the PR into non breaking changes that really only affect docstrings + potentially add TODOs and address follow ups in additional PRs.

PR #167, opened 2026-01-28
Updated draft for the volcanoplot function

vbrennsteiner:
_"After archiving the initial volcanoplot branch, this is the new and improved volcanoplot branch"_

### This PR adds a volcanoplot wrapper

By popular and LLM demand, there should be a volcanoplot wrapper function in alphatools. As such, it should abstract the following functionalities:

- scatterplot
- adding lines to the scatterplot
- adding anchored labels to the scatterplot
- adding a legend
- setting plot x/y limits

### Main difference to the previous PR

Previously, layering of plot features was handled by `volcano()`, which created a suboptimal entanglement of general purpose plotting functionality (layering plots) and the specific application of a volcanoplot. Currently, two new methods + helpers have been added to the plots.py module:

- `layered_plot`: Calls a plotting function repeatedly for any number of specified data layers, ensuring that indices are only plotted once and that indices not used by any layer are plotted in a default color. It operates using a `PlotConfig` instance and a `Callable` capable of interpreting the parameters from the `PlotConfig` and layer-specific parameters. An example is added in `tutorials/tutorial_02_basic_plotting_workflow.ipynb`. 

```
color_dict = {
    "upregulated": BaseColors.get("red"),
    "downregulated": BaseColors.get("blue"),
}
```

To avoid having to specify basic parameters for each layer, we can summarize them in a PlotConfig instance with flexible attributes to accomodate different plotting functions:
```
plot_config = pl.make_scatter_config(
    data=adata,
    x_column="x",
    y_column="y",
    scatter_kwargs={"alpha": 0.7, "s": 50},  # add any kind of coloring/marker specification that should apply to all layers
)
```

The main specification of 'layered_plot' is the `layers` list, which consists of tuples with 3-4 elements:
1. Which column in the data to use for filtering (here 'diff_exp_status')
2. Which value to match in that column (single value or list) (here 'upregulated' for the first and 'downregulated' for the second layer)
3. Which key to look up in the `color_dict` for that layer (here synonymous to the layer match values in 2.)
4. Optional: kwargs dict for that layer (here, upregulated points should be triangular and slightly larger)
```
plot_layers = [
    ("diff_exp_status", "upregulated", "upregulated", {"marker": "^", "s": 100}),
    ("diff_exp_status", "downregulated", "downregulated"),
]
```

The function is called like this on an `AnnData` or `DataFrame` object:
```
fig, axm = create_figure(1, 1, figsize=(6, 6))
ax = axm.next()
Plots.layered_plot(
    ax=ax,
    base_config=plot_config,
    layers=plot_layers,
    color_dict=color_dict,
)
```

### New `volcano()`:

Using `layered_plot` internally, the `volcano` function is drastically simplified and can be called as demonstrated in `tutorials/tutorial_04_volcanoplot.ipynb`. It also incorporates labelling of different layers, which can be specified by their `color_dict` key. `volcano` internally constructs a `PlotConfig` instance to handle default arguments, meaning that users do not have to interact with `PlotConfig`.

As before, the first element refers to the column in data to be used for filtering, the second element specifies a match value or list for that column and the third provides the color lookup for that layer.

```
pois = ["P10291", "P10292", "P10293", "P10294", "P10295"]
plot_layers = [
    ("id", pois, "POI_hypothesis"),
    ("diff_exp_status", "upregulated", "upregulated"),
    ("diff_exp_status", "downregulated", "downregulated"),
]
```

```
color_dict = {
    "POI_hypothesis": BaseColors.get("purple", lighten=0.7), 
    "upregulated": BaseColors.get("orange"),
    "downregulated": BaseColors.get("blue"), 
}
```

```
label_layers = [
    "POI_hypothesis",  # label only the points whose layer points to the "POI_hypothesis" color-key
]
```

Generating the volcanoplot (with anchored labels):
```
Plots.volcano(
    data=adata,
    x_column="log2fc",
    y_column="neg_log10pval",
    color_dict=color_dict,
    layers=plot_layers,
    label_layers=label_layers,
    x_label_anchors=[-3.5, 3.5],
)
```

To create something like this:
image

### Open points:

- Pending the merge of https://github.com/MannLabs/alphapepttools/pull/164, the `label_plot` method will take `AnnData`/`DataFrame` instances with specified columns instead of separate arrays.
- The capabilities of `layered_plot` theoretically allow for layering arbitrary `Callable` instances (plotting functions) with arbitrary global and layer-specific arguments. Perhaps there is a good use case for that in future demonstration notebooks?


src/alphapepttools/pl/plots.py
```diff
@@ -673,8 +1031,11 @@ def scatter(
         legend: str | mpl.legend.Legend | None = None,
         scatter_kwargs: dict | None = None,
         legend_kwargs: dict | None = None,
-        xlim: tuple[float, float] | None = None,
-        ylim: tuple[float, float] | None = None,
+        figure_kwargs: dict | None = None,
+        xlims: tuple[float, float] | None = None,
```
lucas-diedrich: This is a breaking change - Is it necessary to rename the arguments in this PR?

src/alphapepttools/pl/plots.py
```diff
@@ -726,6 +1087,8 @@ def scatter(
             Additional keyword arguments for the matplotlib scatter function. By default None.
         legend_kwargs : dict, optional
             Additional keyword arguments for the matplotlib legend function. By default None.
+        figure_kwargs : dict | None, optional
+            Additional keyword arguments for figure creation. By default None.
         xlim : tuple[float, float], optional
```
lucas-diedrich: Please update the argument names in the docstring or do not change the arguments

src/alphapepttools/pl/plots.py
```diff
@@ -1437,3 +1802,311 @@ def plot_pca_loadings_2d(
...
+        data: ad.AnnData | pd.DataFrame,
+        x_column: str,
+        y_column: str,
+        # Core plotting parameters
+        ax: plt.Axes | None = None,
+        layers: list[tuple] | None = None,
+        color_dict: dict[str, str | tuple] | None = None,
+        # Volcano-specific thresholds
+        x_thresholds: tuple | None = (-1, 1),
+        y_thresholds: tuple | None = (-np.log10(0.05),),
```
lucas-diedrich: I would argue that a `None` default defeats the purpose if a real value is set (i.e. either abstract the `-np.log10(0.05)` logic away into the function, or remove the `None` default). I would prefer the explicit default in the function call.

src/alphapepttools/pl/plots.py
```diff
@@ -368,17 +369,109 @@ def add_legend_to_axes(
...
+    (0, 3.3)
+    """
+    series = pd.Series(values)
+
+    if sym == "max":
+        abs_max = max(abs(series.min()), abs(series.max()))
+        left = -abs_max * padding_factor
+        right = abs_max * padding_factor
+    else:
+        left = series.min() * padding_factor
```
lucas-diedrich: [Claude] This assumes that the minimal value of `series` is negative (otherwise the limit is shifted to the right)

lucas-diedrich (commented):
I like the plots that are returned by the function a lot, they look really clean. 

--- 

I commented on some specific things associated with the pull request. 

I marked issues that were identified by Claude with `[Claude]` based on the following prompt:
```
You are a scientific software developer who is putting a lot of thought into user-friendliness of APIs etc. You are working on a software package for the analysis of proteomics data with anndata as data container. Critically review the changes in the current branch by comparing them to the main branch that  aim to implement a convenience function for volcano plots                                                                                                                                                                                                                                                            
```

---
On a general note, I find the plots module increasingly confusing to review:

- the `Plots` class does not provide a unifying config as its config is not used in any of the functions. This in combination with the `@classmethod` pattern means that individual functions would behave equivalently. This would also make the package more aligned with the other scverse packages (that use individual functions). 
- the number of utility functions + implicit contracts (e.g. expected behavior of configs) makes it complicated to understand the logic

src/alphapepttools/pl/plots.py
```diff
@@ -368,17 +369,109 @@ def add_legend_to_axes(
...
+
+    Examples
+    --------
+    >>> values = np.array([1, 2, 3, -2, -1])
+    >>> _get_plot_lims(values, 1.1, sym="max")
+    (-3.3, 3.3)
+    >>> _get_plot_lims(values, 1.1, set_left=0)
+    (0, 3.3)
+    """
+    series = pd.Series(values)
```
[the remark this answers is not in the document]
vbrennsteiner: That would produce nans if any of the values is nan:

```
from alphapepttools.pl.plots import _get_plot_lims

values = [np.nan, 1, 2, 2, 4, 5, 7, 10]

def _get_plot_lims_without_series(
    values: np.ndarray,
    padding_factor: float = 1.1,
    sym: str | None = None,
    set_left: float | None = None,
    set_right: float | None = None,
) -> tuple[float, float]:
    if sym == "max":
        abs_max = max(abs(min(values)), abs(max(values)))
        left = -abs_max * padding_factor
        right = abs_max * padding_factor
    else:
        left = min(values) * padding_factor
        right = max(values) * padding_factor

    if set_left is not None:
        left = set_left
    if set_right is not None:
        right = set_right

    return (left, right)

print(_get_plot_lims(values))
>>> (1.1, 11.0)

print(_get_plot_lims_without_series(values))
>>> (nan, nan)

```



`pd.Series` methods handle nans gracefully, hence the conversion

src/alphapepttools/pl/plots.py
```diff
@@ -535,6 +626,123 @@ def _dict_keys_to_str(
     return {str(k): v for k, v in dictionary.items()}
 
 
+@dataclass(frozen=True)
+class PlotConfig:
```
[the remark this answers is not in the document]
vbrennsteiner: Trying to understand this - how would the dataclass know what valid types are for any kind of plot that might be layered? 
It's main purpose is to be called by methods like `make_scatter_config` (or e.g. `make_barplot_config`, `make_lineplot_config` in the future), and to hold their data & additional arguments so that the same parameter(s) don't have to be repeated over again for each time. 

In the context of this PR, it is used under the hood to pass shared settings from one plotting layer to the next, e.g. enable users to set alpha/marker/s/etc. once upfront and have that applied to all volcanoplot layers. 

See https://github.com/MannLabs/alphapepttools/pull/167#discussion_r3001056703 for reference

src/alphapepttools/pl/plots.py
```diff
@@ -535,6 +626,123 @@ def _dict_keys_to_str(
...
+    This dataclass holds common configuration parameters for different plot types.
+    It allows handling of dynamic fields, which are stored in the _extra attribute.
+
+
+    """
+
+    data: ad.AnnData | pd.DataFrame | None = None
+    _extra: dict | None = None  # Store additional fields
+
+    def __post_init__(self):
```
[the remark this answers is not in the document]
lucas-diedrich: I was rather referring to the `__post_init__` method, but it's ok

docs/notebooks/studies/study_02_peptidomics_pelsa.ipynb
lucas-diedrich, line 1: as far as I see it, the notebook still uses `at.pl.Plots.scatter`

src/alphapepttools/pl/plots.py
```diff
@@ -377,46 +385,107 @@ def add_legend_to_axes(
...
 
     Parameters
     ----------
-    x_values
-        X coordinates for plotting
-    y_values
-        Y coordinates for plotting
-    labels
-        Labels corresponding to each x,y pair
+    x_values : np.ndarray
```
lucas-diedrich: parameters should not contain type hints

src/alphapepttools/pl/plots.py
```diff
@@ -377,46 +385,107 @@ def add_legend_to_axes(
...
+    padding_factor: float = 1.1,
+    sym: str | None = None,
+    set_left: float | None = None,
+    set_right: float | None = None,
+) -> tuple[float, float]:
+    """Calculate plot limits with optional symmetry and padding.
+
+    Parameters
+    ----------
+    values : np.ndarray
```
lucas-diedrich: parameters should not contain type hints

lucas-diedrich (approved):
Checked that
- volcano plot works without passing axes
- reasonable defaults exist.

PR #208, opened 2026-04-02
Reduction plot - accommodate bpca embeddings  

shanibmo:
Thanks to @lucas-diedrich who raised the issue, the plot_pca and all related dim reduction plotting functions can now be executed for bpca as well (using the addition of the `method = "bpca"` arg, default is `method = "pca"`)


vbrennsteiner (approved):
Nice feature! Getting smooth access to BPCA results should help getting users excited about it, it's a great nan-aware feature to bring to users.

PR #209, opened 2026-04-30
Remove `Plots` wrapper class from `plots.py`

vbrennsteiner:
### Purpose of this PR

The Plots class in the plots.py module has no benefit but makes plotting module ergonomics worse. From now on, instead of `apt.pl.Plots.<function>`, all plotting functions are called at the module level via e.g. `apt.pl.<function>`. 

### What was done
- Remove Plots wrapper class in plots.py
- Add deprecation warning shim class Plots() to guide users towards correct updates --> when trying to call e.g. `Plots.scatter()`, users receive a warning that `Plots` is deprecated and the call should be modified accordingly
- Update all notebooks to the correct syntax
- Update API documentation
- Update __init__.py to correct imports


docs/notebooks/studies/study_02_peptidomics_pelsa.ipynb
lucas-diedrich, line 1: Currently, the notebook is parsed with a Keyboard Interrupt error in the data download section - could you rerun it so that it is shown correctly in the docs?

docs/notebooks/studies/study_01_biomarker_csf.ipynb
lucas-diedrich, line 1: Some plotting functions in this notebook are imported from the alphapepttools namespace - in order to be consistent, e.g. with notebook 02 - would it make sense to use `apt.pl.<>` throughout the notebooks instead of `from alphapepttools.pl import <>`?

docs/notebooks/tutorials/tutorial_02_basic_plotting_workflow.ipynb
```diff
@@ -47,10 +47,11 @@
     "import pandas as pd\n",
     "import string\n",
     "\n",
+    "import alphapepttools as apt\n",
     "from alphapepttools.pl import colors\n",
     "from alphapepttools.pl.colors import show_rgba_color_list, BaseColors\n",
     "from alphapepttools.pl.figure import create_figure, label_axes, save_figure\n",
-    "from alphapepttools.pl.plots import Plots, label_plot\n",
+    "from alphapepttools.pl.plots import label_plot\n",
```
lucas-diedrich: Same here - use `apt.pl.label_plot`?

src/alphapepttools/pl/__init__.py
```diff
@@ -2,10 +2,51 @@
 from .figure import create_figure, label_axes, save_figure
 from .plots import (
     PlotConfig,
-    Plots,
+    Plots,  # Kept here for backward compatibility, Plots leads to a deprecation warning.
```
lucas-diedrich: It's actually a raise condition, correct?

lucas-diedrich (approved):
Looks good! Thanks for changing this, feels much more `scverse` like now. 

Question: Do we consistently use `import alphapepttools as apt` now instead of `at`?

lucas-diedrich:
We should also bump the version to `0.2.0` immediately afterwards (breaking change)

vbrennsteiner:
> Looks good! Thanks for changing this, feels much more `scverse` like now.
> 
> Question: Do we consistently use `import alphapepttools as apt` now instead of `at`?

Good point, I guess `apt` is more recognizable, and the acronym fits so I guess we should switch over to `apt` 👍

src/alphapepttools/pl/__init__.py
vbrennsteiner: true!

docs/notebooks/studies/study_01_biomarker_csf.ipynb
vbrennsteiner: yes I agree, if we change it here, we should change it everywhere

docs/notebooks/tutorials/tutorial_02_basic_plotting_workflow.ipynb
vbrennsteiner: Agreed, changed these imports in our notebooks 👍

PR #211, opened 2026-05-04
add grouped CVs and simplify metrics module

vbrennsteiner:
### Purpose of this PR: 
- Add required functionality to current CV calculation 
- gather pooled feature-level metrics and feature-level metrics in `feature_level.py`
- move non-feature-level metrics out of `feature_level.py` into `sample_and_feature_level.py`

### Current issues:
- The `coefficient_of_variation()` function cannot compute coefficients of variation across sample groups (e.g. CV of all features in control / treatment), which is needed to use CVs in the common use case of assessing variability across treament groups (e.g. in the `study_02_Pelsa_peptidomics.ipynb`  notebook). 
- `group_level.py` contains feature-level functions that aggregate to scalar outputs via sample groups, and an import of `_cv` from `feature_level.py`, so both modules are essentially feature_level metrics modules.
- `feature_level.py` contains  feature+sample level qc metrics that operate on `obs` and `var`.

### What was done:
- A `group_column` argument is added to `coefficient_of_variation()`, analogous to how our imputation functions handle sample groups. When `group_column` is specified, the resulting CV dataframe is saved to `.varm` using the existing `key_added` argument.
- `pooled_coefficient_of_variation()` and `pooled_median_absolute_deviation()`, being wrappers of feature-level metrics, are moved to `feature_level.py` removing the need for `group_level.py`.
- Summary metrics, operating on both feature and observation axes, are moved to a separate `sample_and_feature.py` submodule inside metrics, making the nomenclature explicit and preparing for a future 'sample_level' metrics module.

### Adapted use case
The `study_02_pelsa_peptidomics.ipynb` notebook was changed to use the modified `coefficient_of_variation()` function with the `group_column` feature to calculate CVs across treatment & control samples.


src/alphapepttools/metrics/sample_and_feature_level.py
```diff
@@ -0,0 +1,237 @@
...
+    if axis not in ("obs", "var"):
+        raise ValueError(f"axis must be 'obs' or 'var', got '{axis}'")
+
+    if layer is not None and layer not in adata.layers:
+        raise ValueError(f"Layer '{layer}' not found in adata.layers. Available layers: {list(adata.layers.keys())}")
+
+    data = adata.X if layer is None else adata.layers[layer]
+    special_values_mask = detect_special_values(data, verbosity=0)
+
+    if axis == "obs":
```
lucas-diedrich: `np.mean`

src/alphapepttools/metrics/sample_and_feature_level.py
```diff
@@ -0,0 +1,237 @@
...
+        If True (default), modifies adata inplace and adds the result to adata.obs or adata.var.
+        If False, returns the calculated values without modifying adata.
+
+    Returns
+    -------
+    np.ndarray | None
+        If inplace is False, returns the fraction values as an array.
+        If inplace is True, modifies adata inplace and returns None.
+    """
+    if axis not in ("obs", "var"):
```
lucas-diedrich: Should we also support 0/1?

lucas-diedrich (approved):

lucas-diedrich (approved):

src/alphapepttools/metrics/sample_and_feature_level.py
vbrennsteiner: Yes good point, translates neatly over from Pandas / NumPy 👍

src/alphapepttools/metrics/sample_and_feature_level.py
vbrennsteiner: good point!

PR #216, opened 2026-05-06
IO: Add column mappings in `read_pg_table`

lucas-diedrich:
Previously, when customizing the column mapping in the `read_pg_table` function via the `column_mapping` argument, users had to specify _all_ columns that they wanted to parse. 

This was due to the fact that the `column_mapping` argument was passed directly to the `alphabase.pg_reader.pg_reader_provider.get_reader` call.  Internally, it gets passed to the `__init__` call of the respective `pg_reader` overwriting all defaults. 

This PR adds the argument `additional_column_mapping` to the function signature. Internally this calls the `reader.add_column_mapping` method, which _adds_ the specified column mappings to the default column mapping, while maintaining defaults. 

```python
# Spectronaut reports can contain custom columns, they might be missing in the alphabase default mapping
spectronaut_path = ...
apt.io.read_pg_table( 
    spectronaut_path, 
    search_engine="spectronaut", 
    additional_column_mapping={"new_name": "name_in_pg_table"}
)
```

The arguments `column_mapping` and `measurement_regex` get downgraded to **reader_provider_kwargs, but remain documented. (this does not affect the function signature)


vbrennsteiner (approved):
LGTM! Thanks for catching that, I remember this was a similar situation with the `additional_columns` argument, good to establish proper handling of defaults vs. user inputs 👍

PR #217, opened 2026-05-07
Feature: groupwise normalization

lucas-diedrich:
Batches with very different summed intensities/missingness patterns might need to be normalized separately (e.g. in this DDA/TMT experiment)

> [Function](https://github.com/CharlieBarker/Melanoma-Resistance/blob/deeb753da5c4669492545d5860c64c36526b4bed/src/functions/SLN_normalisation.R); [Analysis script](https://github.com/CharlieBarker/Melanoma-Resistance/blob/deeb753da5c4669492545d5860c64c36526b4bed/src/phosphoproteomic_preprocessing/peptide_prep.R#L57)

This PR adds the option to normalize separately _per batch_. Default behavior remains unchanged.


src/alphapepttools/pp/_utils.py
```diff
@@ -0,0 +1,67 @@
...
+    -------
+    bool
+        True if data contains no missing values, False otherwise
+    """
+    return not np.any(np.isnan(data))
+
+
+def _raise_on_nan_values(
+    data: np.ndarray | pd.DataFrame | pd.Series,
+    mode: Literal["any", "all"] = "any",
```
vbrennsteiner, line 27: Coming to think about it, it's actually not that intuitive to map 'any' and 'all' in this context - how about e.g. `Literal["no_nans", "no_empty_features"]`? It would make it clear that one is concerened with nans in general and the other explicitly deals with feature level missingness

src/alphapepttools/pp/norm.py
```diff
@@ -137,6 +140,11 @@ def normalize(
             total sample intensity is equal to the mean of the total sample intensities across all samples
             - *total_median* The intensity of each feature is adjusted by a normalizing factor so that the
             total sample intensity is equal to the median of the total sample intensities across all samples
+    group_column
+        Column name in `adata.obs` defining groups for group-wise normalization.
+        If `None` (default), computes statistics across all samples.
+        If specified, computes statistics separately for each group.
+        This is useful when working with data from different batches with vastly different intensity distributions.
```
vbrennsteiner: 'vastly different average intensities?'

tests/pp/test_norm.py
```diff
@@ -117,6 +117,28 @@ def test_normalize_function_default(self, strategy, data_type, test_data_factory
         assert len(adata.obs.columns) == 0
         assert len(adata.layers) == 0
 
+    @pytest.mark.parametrize("strategy", ["total_mean", "total_median"])
+    @pytest.mark.parametrize("data_type", ["all_equal", "different", "nan"])
+    def test_normalize_function_group_column(self, strategy, data_type, test_data_factory) -> None:
```
vbrennsteiner, line 122: does this actually hit the group_column case? It seems that the test_data_factory is creating the data once and the ad.concat is using the same data thrice, so it wouldn't really matter if group_column is ignored?

vbrennsteiner (approved):
LGTM - with two considerations:
- let's think about the 'any'/'all' syntax and how we could maybe put in a change now to make the wording more intuitive
- for the group_column test, whether it sufficiently hits the actual case with differerent average intensities in each batch

src/alphapepttools/pp/_utils.py
lucas-diedrich: In my opinion, `any` and `all` is pretty intuitive (and shorter), as it directly aligns with the `pandas` and `numpy` syntax

lucas-diedrich: @mschwoer

src/alphapepttools/pp/norm.py
lucas-diedrich: fixed (removed)

tests/pp/test_norm.py
lucas-diedrich: Thanks, good catch!

I updated the test so that the arrays are now scaled group-wise with a constant scaling factor ($X_1 = X \cdot 1, X_2 = X \cdot 2, X_3 = X\cdot 3, ...$). The expected normalized values per group are now shifted by a constant factor $1, 2, 3, ...$ and different from the expected values of a  global normalization strategy (would be 2x the expected array for all values).

src/alphapepttools/pp/_utils.py
lucas-diedrich: Decision: Leave as is (see #224; )

PR #218, opened 2026-05-07
Feature/filter_data_completeness `any` group 

lucas-diedrich:
The current  function `apt.pp.filter_data_completeness` supports groupwise completeness filtering. The previous logic allows users to filter for features that have a minimal completeness/maximal missingness in _all_ groups (logical AND). However, in many clinical cases, features are specific to a certain condition (these are the most interesting candidates). Removing them is undesired. 

This PR implements the logical OR operation for groupwise completeness by adding the `keep_strategy` parameter. 

```
apt.pp.filter_data_completeness(adata, max_missing=0.5, group="condition", keep_strategy="any")
# keeps all features that are at least 50% complete in at least 1 condition (logical or)

apt.pp.filter_data_completeness(adata, max_missing=0.5, group=None)
# keeps all features that are at least 50% complete across the whole study

# old behavior, new default
apt.pp.filter_data_completeness(adata, max_missing=0.5, group="condition", keep_strategy="all")
# keeps all features that are at least 50% complete in all conditions (logical and)
# e.g. a feature that is found in 0/5 samples in condition 1 and 995/995 samples in condition 2 would be removed
# despite being 99.5% complete across the whole study
```


vbrennsteiner:
> The current function `apt.pp.filter_data_completeness` supports groupwise completeness filtering. The previous logic allows users to filter for features that have a minimal completeness/maximal missingness in _all_ groups (logical AND). However, in many clinical cases, features are specific to a certain condition (these are the most interesting candidates). Removing them is undesired.
> 
> This PR implements the logical OR operation for groupwise completeness by adding the `keep_strategy` parameter.
> 
> ```
> apt.pp.filter_data_completeness(adata, max_missing=0.5, group="condition", keep_strategy="any")
> # keeps all features that are at least 50% complete in at least 1 condition (logical or)
> 
> apt.pp.filter_data_completeness(adata, max_missing=0.5, group=None)
> # keeps all features that are at least 50% complete across the whole study
> 
> # old behavior, new default
> apt.pp.filter_data_completeness(adata, max_missing=0.5, group="condition", keep_strategy="all")
> # keeps all features that are at least 50% complete in all conditions (logical and)
> # e.g. a feature that is found in 0/5 samples in condition 1 and 995/995 samples in condition 2 would be removed
> # despite being 99.5% complete across the whole study
> ```

Very nice, this is really important to support two very opposite use cases: 

- In highly imbalanced settings like clinical studies of rare diseases we need to ensure that a feature is proportionately present in both classes. 
- In pulldown settings or genetic perturbations it's possible that a feature drops below LOD in treatment conditions, so filtering on all groups removes potentially informative features.

With this PR we cover both use-cases 👍

src/alphapepttools/pp/data.py
```diff
@@ -927,33 +928,40 @@ def filter_data_completeness(
...
     Parameters
     ----------
+    adata
+        AnnData object
     max_missing
-        Maximum fraction of missing values allowed. Compared with the fraction of missing values
-        in a "greater than" fashion, i.e. if max_missing is 0.6 and the fraction of missing values
-        is 0.6, the sample or feature is kept. Greater than comparison is used here since the
-        missing fraction may be 0.0, which equals filtering for 100 % data completeness.
+        Maximum fraction of missing values allowed to pass filtering.
```
vbrennsteiner: I would keep an explanation of this parameter independent of the group-handling strategy, it's a bit opaque (and we may want to change it at some point) and an explanation that it works in a "greater than"* fashion and 0.0 filters for 100 % completeness would not go amiss here..

src/alphapepttools/pp/data.py
```diff
@@ -989,54 +997,67 @@ def filter_data_completeness(
...
+    Groupwise filtering — `keep_strategy` controls how per-group results are combined:
 
     .. code-block:: python
 
-        # Consider missingness only in specific groups
-        adata = filter_data_completeness(adata, max_missing=0.5, group_column="group", groups=["A", "B"], action="flag")
+        # No grouping: keep features ≥50% complete across the whole study
+        apt.pp.filter_data_completeness(adata, max_missing=0.5)
+
+        # Logical AND (default): keep features ≥50% complete in *every* condition.
```
vbrennsteiner: wait is this gteq? The implementation checks for lteq I think, so it would be >50 % here?

src/alphapepttools/pp/data.py
```diff
@@ -989,54 +997,67 @@ def filter_data_completeness(
...
-    # Calculate missingness for each group
-    drop_mask = np.array([False] * adata.shape[1])
-    for indices in group_indices.values():
-        missing_fraction = np.isnan(adata[indices, :].X).mean(axis=0)
-        drop_mask |= missing_fraction > max_missing
-        n_dropped = drop_mask.sum()
+        # Aggregate to decision
+        # - any: Any group passes
+        # - all: All groups pass
+        keep_mask = keep_mask.any(axis=0) if keep_strategy == "any" else keep_mask.all(axis=0)
```
vbrennsteiner, line 1044: 👍 for the shift to a keep-mask

vbrennsteiner (approved):
LGTM with the one point: The test for `max_missing = 0.0` still passes so I assume this works, but worth checking whether greater than / greater than or equal is still correctly handled with respect to the documentation.

PR #219, opened 2026-05-08
FEATURE_LEVEL_CONFIG in reader_columns.py has a mistaken singular key

vbrennsteiner:
fix typo in key which breaks the retrieval of precursor level columns by `read_psm_table`: when calling `read_psm_table` wtih level="precursors" and no other column names specified the function crashes with `ValueError: intensity_column is required but not provided and no default found for reader_type='diann' and level='precursors'.` --> that's because the FEATURE_LEVEL_CONFIG in `reader_columns.py` has the singular "precursor" and not "precursors", which this PR fixes.


lucas-diedrich (approved):
lol, yes! just ran into this as well

PR #220, opened 2026-05-14
Fix/combat

lucas-diedrich:
Expose `covariates` argument in combat function. Needed to correctly retain biological variation.

### DIFF
--- a/src/alphapepttools/pp/batch_correction.py
+++ b/src/alphapepttools/pp/batch_correction.py
@@ -134,7 +134,7 @@ def scanpy_pycombat(
         Name of the batch feature in obs, the variation associated with this feature will be corrected.
         Missing values in this column will be replaced by one single "NA" batch
     covariates
-        Columns in `adata.obs` indicating real biological variation. Variation associated with these columns will be retained.
+        Columns in `adata.obs` indicating biological variation. Variation associated with these columns will be retained.
     layer
         Name of the layer to batch correct. If None (default), the attribute adata.X is used
     copy
--- a/src/alphapepttools/pp/batch_correction.py
+++ b/src/alphapepttools/pp/batch_correction.py
@@ -115,7 +115,7 @@ def drop_singleton_batches(
 
 
 def scanpy_pycombat(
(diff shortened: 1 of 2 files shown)


src/alphapepttools/pp/batch_correction.py
```diff
@@ -114,7 +114,9 @@ def drop_singleton_batches(
     return adata
 
 
-def scanpy_pycombat(adata: ad.AnnData, batch: str, layer: str | None = None, *, copy: bool = False) -> ad.AnnData:
```
mschwoer, line 117: just noting: if the signature had been
`def scanpy_pycombat(adata: ad.AnnData, *, batch: str, layer: str | None = None, copy: bool = False) -> ad.AnnData:` ("`*`" moved to left)
this would not be a breaking change ;-)

src/alphapepttools/pp/batch_correction.py
```diff
@@ -131,6 +133,8 @@ def scanpy_pycombat(adata: ad.AnnData, batch: str, layer: str | None = None, *,
     batch
         Name of the batch feature in obs, the variation associated with this feature will be corrected.
         Missing values in this column will be replaced by one single "NA" batch
+    covariates
+        Columns in `adata.obs` indicating real biological variation. Variation associated with these columns will be retained.
```
mschwoer: is "real biological" something different than just "biological"?

maybe ".. indicating biological (as opposed to technical) variation.."

mschwoer (approved):

src/alphapepttools/pp/batch_correction.py
vbrennsteiner: Could do without it imo, in essence it's what the user assumes is real, no way to know if it is..

lucas-diedrich: removed `real`

src/alphapepttools/pp/batch_correction.py
lucas-diedrich: yes we should make this consistent across the package.

vbrennsteiner (approved):
LGTM! This extends the functionality and is in line with scanpy 👍

PR #223, opened 2026-05-24
expose AxisManager in pl init

vbrennsteiner:
makes it possible to use apt.pl.AxisManager in type hints

### DIFF
--- a/src/alphapepttools/pl/__init__.py
+++ b/src/alphapepttools/pl/__init__.py
@@ -1,5 +1,5 @@
 from .colors import BaseColormaps, BaseColors, BasePalettes, MappedColormaps, get_color_mapping, show_rgba_color_list
-from .figure import create_figure, label_axes, save_figure
+from .figure import AxisManager, create_figure, label_axes, save_figure
 from .plots import (
     PlotConfig,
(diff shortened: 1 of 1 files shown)


lucas-diedrich (approved):
Don't really get the reasoning - this already works 

```python
from alphapepttools.pl.figure import AxisManager

def func(manager: AxisManager) -> None:
    ... 
```

Note the full path (including module)

vbrennsteiner:
> Don't really get the reasoning - this already works
> 
> ```python
> from alphapepttools.pl.figure import AxisManager
> 
> def func(manager: AxisManager) -> None:
>     ... 
> ```
> 
> Note the full path (including module)

Just to stay in line with the emerging pattern of importing `apt` in the beginning of a notebook and then referencing the type as `apt.pl.AxisManager`. Users might be confused why the all of a sudden have to import that one specific class when everything else is in the apt.<module> namespace

lucas-diedrich:
I'd argue that this is an implementation detail and would not make it a "truly" public API to keep the exposed surface minimal (what if we want to change the interface - do we have to add deprecation warnings etc so that we don't break anyone else's code)

lucas-diedrich:
@mschwoer what do you think?

PR #226, opened 2026-06-02
Remove plot index offset

vbrennsteiner:
### Purpose of this PR

`_extract_groupwise_plotting_data()` prepares data for plotting with box/violin/bar plotting functions. Until now, it assigned x-axis positions in a 1-indexed manner. The problem with this is that combinations of `alphapepttools` functions with e.g. Seaborn implementations on the same ax object run into off-by-one errors, since comparable Seaborn plots are 0-indexed. This PR fixes alphapepttools box/violon/bar plots to 0-indexing.


lucas-diedrich (approved):
Good call!

PR #231, opened 2026-06-18
Adjust fonts and lines

vbrennsteiner:
### Minor adjustment to create more paper-ready figures:

Recap: `create_figure` comes with Nature-style presets for the `figsize` argument: when calling `create_figure(figsize=("1", "1"))`, it creates a panel measured out to one column of a typical Nature-style figure panel.

### What was fixed:

Considering these presets, default small, medium and large font sizes and line widths were too large, clashing with the intended appearance on pages & in print. This PR adjusts font sizes and linewidths and fixes some inconsistencies where dimensions were not properly retrieved from the `defaults.py` module.

### DIFF
--- a/src/alphapepttools/pl/plots.py
+++ b/src/alphapepttools/pl/plots.py
@@ -167,7 +167,7 @@ def add_lines(
     linestyle
         Line style (e.g., `"--"`, `"-"`, `":"`)
     linewidth
-        Line width, defaults to `config["linewidths"]["medium"]`
+        Line width, defaults to `config["linewidths"]["large"]`
     line_kwargs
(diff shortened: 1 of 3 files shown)


src/alphapepttools/pl/plots.py
```diff
@@ -195,6 +195,9 @@ def add_lines(
         raise ValueError("linetype must be 'vline' or 'hline'")
     line_func = ax.axvline if linetype == "vline" else ax.axhline
 
+    if linewidth is None:
```
lucas-diedrich: Wouldn't it be more transparent to surface the default value to the function call? 

(Not blocking)

lucas-diedrich (approved):

PR #237, opened 2026-06-19
`metrics` test expansion

vbrennsteiner:
Add tests for helper functions in metrics


mschwoer (approved):
LTMG (Les tests me goûtent.)

PR #239, opened 2026-06-26
Type hints

mschwoer:
Add type hints and `ty` to pre-commit hooks.

Note that there's an awful lot of exceptions still, will revisit them in a follow-up PR.


pyproject.toml
```diff
@@ -191,6 +191,43 @@ omit = [
...
+# backend (ndarray | csr_matrix | AwkArray | ZarrArray | CSRDataset | None | ...).
+# The pipeline treats these as dense numpy arrays (it asserts no-NaN and does
+# `data[idx, :] = ...`), so subscripting and in-place assignment are safe at
+# runtime but cannot be proven to a static checker.
+not-subscriptable = "ignore"
+invalid-assignment = "ignore"
+# Same union flows into helper calls expecting `np.ndarray`; in this stack the
+# rule only ever fires on this union and on external libraries whose stubs
+# declare `str` where the code passes `Path`, so it carries no real signal here.
+invalid-argument-type = "ignore"
```
lucas-diedrich, line 219: This is one of the key rules in `mypy`, don't we miss a lot of signal when this is ignored?

pyproject.toml
```diff
@@ -191,6 +191,43 @@ omit = [
...
+
+[tool.ty.rules]
+# TODO revisit these exclusions
+# anndata types `.X`/`.layers[...]` as a 12+ member union of every storage
+# backend (ndarray | csr_matrix | AwkArray | ZarrArray | CSRDataset | None | ...).
+# The pipeline treats these as dense numpy arrays (it asserts no-NaN and does
+# `data[idx, :] = ...`), so subscripting and in-place assignment are safe at
+# runtime but cannot be proven to a static checker.
+not-subscriptable = "ignore"
+invalid-assignment = "ignore"
```
lucas-diedrich, line 215: This is one of the key rules in mypy, don't we miss a lot of signal when this is ignored?

pyproject.toml
```diff
@@ -191,6 +191,43 @@ omit = [
...
+# cannot see are imported.
+possibly-missing-submodule = "ignore"
+# The remaining rules below fire only on this stack's untyped/dynamic libraries
+# (cmcrameri's generated colormaps, patsy, scanpy/numpy/pandas overloads, the
+# `.X` union) and on the optional `alphaquant` dependency. Genuine issues these
+# surfaced were already fixed in code; disabling them keeps `ty check src` clean
+# so it can run as a pre-commit gate that still enforces every other ty rule.
+unresolved-attribute = "ignore"
+no-matching-overload = "ignore"
+invalid-return-type = "ignore"
```
lucas-diedrich: This is one of the key rules in mypy, don't we miss a lot of signal when this is ignored?

lucas-diedrich (commented):
Some of the settings seem to be pretty permissive/miss key functionality of mypy? 

I asked [Claude](https://claude.ai/share/4cf10775-792d-45d7-a91b-bcefd796f22e) which bugs would be missed at the moment:
```
invalid-argument-type: np.log2(adata) instead of np.log2(adata.X); passing adata.obs (a DataFrame) to a function wanting np.ndarray; swapping two same-looking positional args.
invalid-assignment: self.intensities: np.ndarray = adata; n_peptides: int = adata.var_names.
invalid-return-type: a def normalize(...) -> np.ndarray that hits a branch with no return (returns None), or returns adata instead of adata.X.
unresolved-attribute: adata.obms["X_pca"] (typo), result.pval when it's result.pvalue, accessing an attribute you renamed.
unresolved-import: from scnapy import ..., or forgetting to declare a newly-used dependency.
no-matching-overload/unsupported-operator/not-iterable: pd.concat with the wrong arg shape; None * scale_factor when an optional wasn't set; iterating an Optional you assumed was populated.
```

pyproject.toml
mschwoer: yes, this required larger code changes, see follow-up PR(s)

pyproject.toml
mschwoer: see https://github.com/MannLabs/alphapepttools/pull/243

mschwoer:
@lucas-diedrich could you have another look?
also, how shall we deal with the failing PRE-RELEASE test? I think these failures are unrelated to the code changes .. still merge and fix in a dedicated PR?
https://github.com/MannLabs/alphapepttools/actions/runs/31085825306/job/92564907334?pr=239

lucas-diedrich:
Will have a look in a second 
Yes I'd suggest that, mudata apparently changed their settings API 

- https://github.com/scverse/mudata/blob/2c2fe7501695b460215b7bc1be9d6b5ced383227/CHANGELOG.md?plain=1#L18

Fix
```python
# old mudata.set_options(pull_on_update=true)
mudata.settings.pull_on_update = True
```

src/alphapepttools/metrics/feature_level.py
```diff
@@ -205,7 +205,7 @@ def _set_nested_dict(
 
 def _compute_pooled_groupwise_metric(
     adata: ad.AnnData, func: Callable[[np.ndarray], float], group_column: str, layer: str | None = None, **kwargs
-) -> dict[str, list[float]]:
+) -> dict[Hashable, float]:
```
lucas-diedrich, line 208: today I learned the most general representation of a key in python dictionaries

src/alphapepttools/pl/colors.py
```diff
@@ -519,7 +519,7 @@ def get(
         color_name: str | tuple,
         lighten: float | None = None,
         alpha: float | None = None,
-    ) -> tuple:
+    ) -> tuple | str:
```
lucas-diedrich, line 522: [not in this pr] this seems like an orange flag that the color can be returned either as tuple or string

lucas-diedrich (approved):
looks good, thanks for addressing the comments

src/alphapepttools/pl/colors.py
mschwoer: you mean an `"orange"` or an `("orange",)` flag?

PR #240, opened 2026-06-26
Expanded ebayes ttest

vbrennsteiner:
### Purpose of this PR

The existing implementation of `diff_exp_ebayes()` works well for differential expression data but has some critical shortcomings, the first of which is directly caused by the underlying `lmFit()` implementation of `inmoose`:

1. No handling of NaN-containing data, i.e. requiring 100 % feature completeness to avoid crashing `lmFit()`
2. Only computing one contrast (e.g. Compound_X vs. Control) at a time
3. Not supporting any linear covariates that could be added to the linear fit to explain batch variance

`ebayes_expanded.py` addresses all of these points in its new version of `diff_exp_ebayes`, which consists of six new custom functions to sidestep all `inmoose` dependencies except `limma.ebayes` itself. Its critical component is `nan_lmfit()`, performs the same kind of linear fitting as `lmFit()` but tolerates missing values:

- `build_design_matrix` to construct a 1/0 matrix for fitting the linear coefficients
- `generate_feature_mask` to construct a selection mask for features with sufficient control samples (some tests experimental setups require an up-front check to only assess features with sufficient control replicates. This function separates that logic from the actual linear fitting below
- `nan_lmfit` to perform the actual feature-wise fitting while tolerating missing features. Can optionally take a `feature_mask` from `generate_feature_mask`
- `make_contrasts` constructs a contrasts matrix to construct fold changes from the coefficient's delta (i.e. Beta_treatment - Beta_control) results in the respective fold-change
- `run_contrasts` Actually compute the fold changes, i.e. the t-test but keep unscaled variance and standard deviation in a separate vector; Since the coefficients (i.e. also the coefficient's delta's) standard error consists of the standard deviation and the variance, and empirical Bayes moderation adjusts the standard deviation, these two must remain separate here
- `ebayes_moderation` takes the coefficients and the unscaled variance and standard deviation arrays and performs empirical Bayes moderation. This results in modified t-values and p-values, and is also the only point where we still need `inmoose`

These scripts are coordinated in the new version of `diff_exp_ebayes`, which can now handle nans, multiple contrasts and linear covariates, i.e. to model batch effects.

### Critical tests

The key test is numerical equivalence between the new and old `diff_exp_ebayes` versions in `test_diff_exp.py` on a vanilla dataset without missing values or covariates and with only one contrast. Furthermore, all new functions are covered by tests, with special attention on `nan_lmfit` to ensure that there are no index misalignments of features or coefficients when arrays are brought back to full shape after fitting with filtered design matrices that avoid missing samples for the respective feature

### Outlook

The new function can be used largely as-is in place of the existing `diff_exp_ebayes` with one breaking change: results are now returned in a {contrast_name: dataframe} structured dict, rather than a simple dataframe in the previous single-comparison case. This is a direct result of supporting multiple contrasts now.

Once a port of the `limma.ebayes` function from `inmoose` exists, `inmoose` can be dropped as a dependency entirely.

### Key features of the new implementation compared to the old diff_exp_ebayes:

- **Every feature is returned.** The previous implementation silently dropped any feature with a
  missing value. This one fits all features and returns all of them, setting `log2fc`, `p_value`,
  `fdr` and `stat` to NaN only where the replicate gate suppresses them. Filtering on
  significance still behaves as expected (NaN comparisons are False), but row counts and
  index-based joins change.
- **The variance prior is estimated from all fittable features**, not only complete ones, so
  p-values differ for otherwise identical input unless the data is filtered upstream with
  :func:`alphapepttools.pp.filter_data_completeness`.
- **Results are stacked over contrasts**, one row per (feature, contrast) pair, so the index
  repeats for a multi-contrast call. Filter on `condition_pair` to recover a single contrast.
- **`B` and `AveExpr` are no longer returned**; `stat` still is.


src/alphapepttools/pp/data.py
```diff
@@ -923,6 +927,35 @@ def _validate_adata_for_completeness_filter(adata: ad.AnnData, action: str, var_
         )
 
 
+def _validate_max_missing(
```
mschwoer: this is https://github.com/MannLabs/alphapepttools/pull/242, right?

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,590 @@
...
+    _HAS_INMOOSE = False
+
+from alphapepttools.tl.stats import nan_safe_bh_correction
+from alphapepttools.tl.utils import (
+    determine_max_replicates,
+    negative_log10_pvalue,
+)
+
+
+def build_design_matrix(
```
mschwoer: could this be private? (check also the rest of the public API of this module)

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,590 @@
...
+    for covariate in covariate_names:
+        dm.loc[adata.obs[covariate_column] == covariate, covariate] = 1
+
+    condition_col_idxs = {name: dm.columns.get_loc(name) for name in condition_names}
+    covariate_col_idxs = {name: dm.columns.get_loc(name) for name in covariate_names}
+
+    return dm, {"condition_col_idxs": condition_col_idxs, "covariate_col_idxs": covariate_col_idxs}
+
+
+def summarize_design_matrix(
```
mschwoer: `print_design_matrix_summary()` ?

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,590 @@
...
+    # Validate the A conditions (comparison[0])
+    a_conditions = comparison[0]
+    if a_conditions == "_ALL_":
+        a_conditions = [level for level in between_levels if level != b_condition]
+    elif isinstance(a_conditions, str):
+        a_conditions = [a_conditions]
+
+    for a_condition in a_conditions:
+        if a_condition not in between_levels:
+            raise ValueError(f"Condition '{a_condition}' not found in column '{between_column}'.")
```
mschwoer: this method is quite long, move this to a dedicated method `_validate_input()` or smth?

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,590 @@
...
+                "p_value": p_values,
+                "-log10(p_value)": neg_log10_pvalues,
+                "fdr": fdr_pvalues,
+                "-log10(fdr)": neg_log10_fdr,
+                "method": "limma_ebayes_inmoose_expanded",
+                "max_level_1_samples": max_level_1_samples,
+                "max_level_2_samples": max_level_2_samples,
+            },
+            index=adata.var_names,
+        )
```
mschwoer: this has a different level of abstraction than the rest of this method's code, move it to a dedicated method

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,590 @@
...
+    if return_coefficients:
+        # rows of B are design-matrix columns (conditions + covariates), in column order
+        coefficients = pd.DataFrame(
+            lm_fit["B"].T,
+            index=adata.var_names,
+            columns=design_matrix.columns,
+        )
+        return results, coefficients
+
+    return results
```
mschwoer: `return results, []` or `return results, None` here in order to keep the methods return signature unconditional

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,590 @@
...
+    results = {}
+    for contrast_idx, contrast_name in enumerate(contrast_names):
+        # By convention the contrast is named "A_VS_B" (comparison[0] vs comparison[1]).
+        a_level, b_level = contrast_name.split("_VS_")
+
+        p_values = ebayes_results["p"][contrast_idx].copy()
+        log2fc = contrast_results["log2fc"][contrast_idx].copy()
+
+        # Replicate gate: suppress the fold change unless both conditions have enough observed values
+        keep = np.ones(adata.n_vars, dtype=bool)
```
mschwoer: (nit) `keep_mask`

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,590 @@
...
+
+    # Step 5: Extract contrasts and write output DataFrame with standardized columns for each contrast
+    contrast_names = contrasts_from_matrix(contrast_matrix, b_condition)
+    if len(contrast_names) != contrast_results["log2fc"].shape[0]:
+        raise ValueError("Number of contrast names does not match number of contrasts in results.")
+
+    results = {}
+    for contrast_idx, contrast_name in enumerate(contrast_names):
+        # By convention the contrast is named "A_VS_B" (comparison[0] vs comparison[1]).
+        a_level, b_level = contrast_name.split("_VS_")
```
mschwoer: introduce a string constant for `_VS_` to make the coupling in the code explicit

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,590 @@
...
+        live_col_idxs = np.where(live_col_mask)[0]
+
+        # Fit linear model using generic least squares solver
+        B_live, _residuals, rank, _s = np.linalg.lstsq(X_live, y_obs, rcond=None)
+
+        # Degrees of freedom based on current non-missing observations and design matrix rank
+        m = y_obs.shape[0]
+        df = m - rank
+
+        # DF check: if df <= 0, skip this feature
```
mschwoer: please check for redundant comments

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,590 @@
...
+    Returns
+    -------
+    tuple[np.ndarray, np.ndarray]
+        p-values and moderated t-statistics for each contrast and feature, both of shape (n_contrasts, n_features). Features that were never fit are returned as NaN.
+
+    """
+    n_contrasts, P = log2fcs.shape
+
+    # Remove features that have no fit at all, i.e. NaN columns in log2fcs, stdevs_unscaled, or nan in sigma2, or dfs
+    valid = np.isfinite(sigma2) & np.isfinite(dfs) & (dfs > 0)
```
mschwoer: `valid_mask`?

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,590 @@
...
+    fit = limma.MArrayLM(
+        coefficients=pd.DataFrame(log2fcs[:, valid].T, index=valid_idx),
+        stdev_unscaled=pd.DataFrame(stdevs_unscaled[:, valid].T, index=valid_idx),
+        sigma=np.sqrt(sigma2[valid]),  # SD, not variance
+        df_residual=dfs[valid],
+        cov_coef=None,
+    )
+    fit = limma.eBayes(fit)
+
+    # scatter back to full (88, P), excluded features stay NaN
```
mschwoer: `(88, P)` ?

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: Agreed, it's now a default separator string set by the defaults.py module 👍

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: leftover from dev on a dataset, fixed to be (n_contrasts, P) 👍

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: Changed that, but keep in mind that there's an upcoming PR to harmonize differential expression return signatures, so it will likely not stay this way for long. Since that harmonization touches all diff_exp modules I didn't want to weave it into this PR which is already quite big..

src/alphapepttools/pp/data.py
```diff
@@ -274,8 +274,12 @@ def _filter_by_dict(
         merged_filter_mask = np.all(filter_masks, axis=0)
     elif logic == "or":
         merged_filter_mask = np.any(filter_masks, axis=0)
+    else:
```
lucas-diedrich, line 279: this is `any` and `all` in other contexts - standardize?

src/alphapepttools/pp/data.py
```diff
@@ -923,26 +931,117 @@ def _validate_adata_for_completeness_filter(adata: ad.AnnData, action: str, var_
         )
 
 
+def _resolve_max_missing(
```
lucas-diedrich, line 972: honestly, I feel like we're still in `alpha` and breaking changes are to be expected - so I would not sweat too much about backwards compatibility and just document breaking changes in the release notes

lucas-diedrich, line 972: Did this leak into this branch from #242

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
lucas-diedrich, line 1: The file name `ebayes_expanded.py` will not age well

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,723 @@
...
+    )
+
+    # Step 5: Extract contrasts and write output DataFrame with standardized columns for each contrast
+    contrast_names = _contrasts_from_matrix(contrast_matrix, b_condition)
+    if len(contrast_names) != contrast_results["log2fc"].shape[0]:
+        raise ValueError("Number of contrast names does not match number of contrasts in results.")
+
+    results = {}
+    for contrast_idx, contrast_name in enumerate(contrast_names):
+        # By convention the contrast is named "A_VS_B" (comparison[0] vs comparison[1]).
```
lucas-diedrich, line 720: (Not this PR): I would suggest to split contrasts into 2 columns in the result: `comparison` (`comparison[0]`) and `baseline` (`comparison[1]`). This makes it easier to parse the comparisons afterwards.

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,723 @@
...
+        p_values = ebayes_results["p"][contrast_idx].copy()
+        log2fc = contrast_results["log2fc"][contrast_idx].copy()
+
+        # Replicate gate: suppress the fold change unless both conditions have enough observed values
+        keep = _replicate_gate_mask(adata, between_column, a_level, b_level, a_min_required, b_min_required)
+        p_values[~keep] = np.nan
+        log2fc[~keep] = np.nan
+
+        # Sample counts per level, mirroring ebayes.py. The name is "A_VS_B".
+        max_level_1_samples, max_level_2_samples = determine_max_replicates(adata, between_column, a_level, b_level)
```
lucas-diedrich, line 732: (not this PR): as discussed, `determine_max_replicates` reports a very unintuitive metric (maximum number of replicates across all features), compared to a more intuitive feature-wise measure.

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,723 @@
...
+        Column name in adata.obs containing linear covariate levels, by default None.
+    a_min_required : int | None, optional
+        Minimum number of observed values required in the A condition (comparison[0]) of each contrast. Per
+        contrast, features with fewer observed values in A have their fold change suppressed (set to NaN) before
+        FDR correction. If None, the A gate is disabled. By default None.
+    b_min_required : int | None, optional
+        Minimum number of observed values required in the B condition (comparison[1]). Per contrast, features with
+        fewer observed values in B have their fold change suppressed (set to NaN) before FDR correction. If None,
+        the B gate is disabled. By default None.
+    return_coefficients : bool, optional
```
lucas-diedrich: you could consider to keep the return signature constant (always return a single dataframe) but add the coefficients as "extra" columns to the `diff_exp` result?

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,723 @@
...
+        )
+
+    if return_coefficients:
+        # rows of B are design-matrix columns (conditions + covariates), in column order
+        coefficients = pd.DataFrame(
+            lm_fit["B"].T,
+            index=adata.var_names,
+            columns=design_matrix.columns,
+        )
+        return results, coefficients
```
lucas-diedrich: return concatenated dataframe instead of dict of data frames? 

```
...
return pd.concat(results)
```

Also, the keys of the dictionary are redundant with the column `contrast_name` in the dataframe

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,723 @@
...
+    pd.DataFrame
+        A design matrix suitable for linear modeling, where columns are features and rows are samples. Each condition is encoded as a separate column. Covariate levels are added as additional columns in a k-1 fashion (one-hot encoding with one level dropped to avoid multicollinearity).
+    dict
+        A dictionary to navigate the design matrix with the following keys:
+        - condition_col_idxs: A dictionary mapping each condition to its corresponding column index in the design matrix.
+        - covariate_col_idxs: A dictionary mapping each covariate level to its corresponding column index in the design matrix. Empty if covariate_column is None.
+
+    """
+    if condition_column not in adata.obs.columns:
+        raise ValueError(f"Condition column '{condition_column}' not found in adata.obs.")
```
lucas-diedrich: `KeyError`

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,723 @@
...
+
+    if adata.obs[condition_column].isna().any():
+        raise ValueError(f"Condition column '{condition_column}' contains NaN values.")
```
lucas-diedrich: `KeyError`

```python
df = pd.DataFrame()
df["non_existent"]
> KeyError: 'non_existent'
```

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,723 @@
...
+    )  # k-1 for covariates
+
+    nrows = adata.n_obs
+    ncols = len(condition_names) + len(covariate_names)
+
+    dm = pd.DataFrame(
+        data=np.zeros((nrows, ncols), dtype=int), index=adata.obs_names, columns=condition_names + covariate_names
+    )
+
+    for condition in condition_names:
```
lucas-diedrich: What about `pd.get_dummies(adata.obs[condition_column], dtype=int)`
https://pandas.pydata.org/docs/reference/api/pandas.get_dummies.html, which gets rid of this logic?

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: This will be part of a future PR focused on unifying the differential expression signatures (https://mannlabs.atlassian.net/browse/AX-1513?atlOrigin=eyJpIjoiZWRhZjIwNjI0YmNlNDNmY2I3M2U2YTI0MGI4MmI3NDEiLCJwIjoiaiJ9)

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: This will be part of a future PR focused on unifying the differential expression signatures (https://mannlabs.atlassian.net/browse/AX-1513?atlOrigin=eyJpIjoiZWRhZjIwNjI0YmNlNDNmY2I3M2U2YTI0MGI4MmI3NDEiLCJwIjoiaiJ9)

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: The intention was to show the maximum number of replicates that exist for a given level. Different question from a feature-wise metric, however I agree this could be useful in the future. For now the main point was to match the existing ebayes.py return since this is a reimplementation.

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: As I see it, separating those begs for a mixup since it is not always semantically clear what the "baseline" and what the "comparison" should be. Any disconnect in their order invites failure modes where it becomes unclear what was compared against what, i.e. you see a point in the volcanoplot moving left and it's no longer clear if the comparison was A vs B or B vs A. I actually experimented with calling the two arms "condition" and "control" but it became clear that sometimes this is not clear-cut and you're just comparing two conditions where neither is a dedicated "control". If that happens, the order must be unambiguously established. Also, results become harder to filter/query when there is no single contrast to search for.

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
lucas-diedrich: I would argue that this argument also holds for "something_vs_something_else" and that a single "vs" string is even less clear on the directionality

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,723 @@
...
+
+    # Step 5: Extract contrasts and write output DataFrame with standardized columns for each contrast
+    contrast_names = _contrasts_from_matrix(contrast_matrix, b_condition)
+    if len(contrast_names) != contrast_results["log2fc"].shape[0]:
+        raise ValueError("Number of contrast names does not match number of contrasts in results.")
+
+    results = {}
+    for contrast_idx, contrast_name in enumerate(contrast_names):
+        # By convention the contrast is named "A_VS_B" (comparison[0] vs comparison[1]).
+        a_level, b_level = contrast_name.split(tl_defaults.CONDITION_SPLIT_STRING)
```
lucas-diedrich, line 721: Splitting the string here is prone to errors (e.g. when a condition is called `DRUG_VS_1` this would fail -> splits to `DRUG`, `1` `CONDITION 2`)- why not retain the contrasts  as strings throughout the function?)

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
lucas-diedrich: I would argue that this argument also holds for "something_vs_something_else" and that a single "vs" string is even less clear on the directionality

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
```diff
@@ -0,0 +1,723 @@
...
+    if not design_matrix.index.equals(adata.obs_names):
+        raise ValueError("Design matrix index does not align with adata.obs_names.")
+
+    # Get general shape information
+    K = design_matrix.shape[1]
+    P = adata.n_vars
+
+    # Convert design matrix and response to numpy arrays
+    X = design_matrix.to_numpy()
+    Y = adata.X
```
lucas-diedrich: so this only works with the X layer?

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
lucas-diedrich, line 1: This module and the function are currently not "user facing" (i.e. exposed in `apt.tl`) - is this intended?

vbrennsteiner: Agreed, it is not going to stick around. Next PR is going to remove the existing ebayes.py module and replace it with the new one.

src/alphapepttools/pp/data.py
vbrennsteiner: in the past the deprecation message for the Plots() class was well received, hence the added guidance

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: Additionally, changing this annotation would impact the remaining differential expression modules which currently operate on the A_VS_B syntax (which is taken from AlphaQuant btw), putting this change out of scope for this PR

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: for now, same for all other diff_exp functions as they are currently implemented - though it would be worthwhile expanding their signatures in the future with a "layer" argument.

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: yeah good point, that is the canonical way to handle dummy variables so it should be used instead of custom logic

src/alphapepttools/pp/data.py
vbrennsteiner: agreed, but not part of this PR so converted to an issue. This only came up here because of the new type checking which flagged the return here.

src/alphapepttools/tl/diff_exp/ebayes_expanded.py
vbrennsteiner: I agree but this is out of scope since this failure mode concerns all other differential expression functions as well. This has to be addressed on the scope of the diff_exp module (https://mannlabs.atlassian.net/jira/software/c/projects/AX/boards/103?selectedIssue=AX-1513). Note that the "_VS_" convention comes from AlphaQuant's "condition_pair" syntax so something along those lines would most likely be needed, but in the refactor I would suggest a checker to flag problematic input strings, which is defensible insofar as users can set the separator string themselves in defaults to avoid "_VS_" or whatever we settle on.

src/alphapepttools/tl/diff_exp/ebayes.py
```diff
@@ -1,239 +1,812 @@
...
-    level_2: str,
-) -> tuple:
-    """Generate Patsy design matrix with correct column ordering.
+def _one_hot(labels: pd.Series, *, drop_first: bool = False) -> pd.DataFrame:
+    """One-hot encode sample labels, keeping columns in order of first appearance.
 
-    Patsy automatically sorts columns alphabetically, which can cause issues with
-    contrast specifications. This function creates a design matrix and returns
-    the column names in the desired order (level_1, level_2).
+    Casting to ``object`` first is what makes this safe for AnnData: ``pd.get_dummies`` on a
```
lucas-diedrich: Technically an implementation detail (=inline comment, not docstring)

src/alphapepttools/tl/diff_exp/ebayes.py
```diff
@@ -1,239 +1,812 @@
...
+
+    Returns
+    -------
+    pd.DataFrame
+        Indicator matrix indexed like ``labels``, with columns ordered by first appearance.
+
+    """
+    labels = labels.astype(object)
+    dm = pd.get_dummies(labels, dtype=int, drop_first=drop_first)
+    # get_dummies sorts its columns; restore order of first appearance, minus any dropped level.
```
lucas-diedrich, line 59: why is this necessary for the downstream code to work?

src/alphapepttools/tl/diff_exp/ebayes.py
```diff
@@ -1,239 +1,812 @@
...
+        # Extract the response vector for the current feature
+        y = Y[:, j]
+
+        # Subset to observed values
+        obs_mask = ~np.isnan(y)
+        y_obs = y[obs_mask]
+        X_obs = X[obs_mask, :]
+
+        # Drop those columns from X_obs which only have zeros after subsetting to obs
+        live_col_mask = X_obs.sum(axis=0) > 0
```
lucas-diedrich: is this $>$ safe to check? Could an `X_obs` be negative and non-zero (e.g when running on scaled values)?

src/alphapepttools/tl/diff_exp/ebayes.py
```diff
@@ -1,239 +1,812 @@
...
+        control_value = row[control_condition]
+        treatment_cols = [col for col in contrast_matrix.columns if col != control_condition and row[col] != 0]
+        if len(treatment_cols) != 1:
+            raise ValueError(
+                f"Each contrast row must have exactly one non-zero treatment column, found {treatment_cols}."
+            )
+        treatment = treatment_cols[0]
+        treatment_value = row[treatment]
+
+        # Name must match the sign of log2fc = C @ B: control=1/treatment=-1 gives
```
lucas-diedrich, line 377: Why can this (-1/1 for treatment/control) vary?

src/alphapepttools/tl/diff_exp/ebayes.py
```diff
@@ -1,239 +1,812 @@
...
+    """
+    fdr_pvalues = nan_safe_bh_correction(p_values)
+
+    df = pd.DataFrame(
+        {
+            "condition_pair": contrast_name,
+            "protein": var_names,
+            "log2fc": log2fc,
+            "p_value": p_values,
+            "-log10(p_value)": [negative_log10_pvalue(p) for p in p_values],
```
lucas-diedrich, line 600: (not this PR) we should refactor this function (`negative_log10_pvalue`) to make it applicable to arrays 

```python
np.clip(-np.log10(value), a_min=None, a_max=tl_defaultls.CEILING_NEGATIVE_LOG10_PVALUE)
```

src/alphapepttools/tl/diff_exp/ebayes.py
```diff
@@ -1,239 +1,812 @@
...
+    KeyError
+        If `between_column` is not in adata.obs, or any condition in `comparison` is not a level of it.
 
     Notes
     -----
-    This implementation removes all features with any missing values to ensure
-    compatibility with the inmoose Limma implementation. If features with missing
-    values should be retained, their missing values must be imputed prior to running
-    this function.
+    This function replaces an earlier complete-cases-only implementation that wrapped
```
lucas-diedrich: Is this historical comment necessary? Should move this to the changelog/release notes

tests/tl/test_diff_exp.py
```diff
@@ -606,76 +615,123 @@ def test__standardize_alphaquant_results(
...
 
-    # The method labels are intentionally distinct between the two implementations.
-    assert original["method"].unique().tolist() == ["limma_ebayes_inmoose"]
-    assert expanded["method"].unique().tolist() == ["limma_ebayes_inmoose_expanded"]
+
+# The multi-contrast path has no counterpart in the old implementation, so the oracle is the only
+# independent check on it: a joint fit over three conditions must place the same fold changes as two
+# separate pairwise limma fits. Only log2fc is comparable -- the joint fit pools residual variance
+# across all three conditions, so its eBayes prior (and hence its p-values) legitimately differs.
+@pytest.mark.skipif(not _HAS_INMOOSE, reason="inmoose not installed")
```
lucas-diedrich, line 715: (not this PR)

Just fyi, I think this is a neat pytest functionality that would reduce the number of skip decorators you'd have to implement

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip


I would also split the diff_exp tests into different modules at some point, as the file is at 1500 LOC now

lucas-diedrich (approved):
great effort, looks good!

src/alphapepttools/tl/diff_exp/ebayes.py
vbrennsteiner: The main comparison is all key-based now, but as some matrices emerge during the computation which don't have column labels I wanted to explicitly fix the order here.

tests/tl/test_diff_exp.py
vbrennsteiner: Good point on the importorskip, I'll look into it - wrt the split, yes at some point that might make sense, it could just be one test.py per function, especially when we add more differential expression methods 👍 (https://mannlabs.atlassian.net/browse/AX-1563?atlOrigin=eyJpIjoiNWIxMDIyYTQ2NzM2NGZlOGI0NWQyNzVjZTE0ZjE3ZGYiLCJwIjoiaiJ9)

src/alphapepttools/tl/diff_exp/ebayes.py
vbrennsteiner: good point, replaced with a more explicit check that captures the "we don't want all-zero columns here" meaning better 👍

src/alphapepttools/tl/diff_exp/ebayes.py
vbrennsteiner: The treatment/control nomenclature here is relative to what the original inputs actually were, determining the order of the comparison string & the direction of fc a downstream volcanoplot.

src/alphapepttools/tl/diff_exp/ebayes.py
vbrennsteiner: this is a good point, moved it to the PR description & will put it in the release notes eventually 👍

PR #241, opened 2026-06-28
[Docs]: spatialdata integration

lucas-diedrich:
Update DVP tutorial with spatialdata plotting 

image


shanibmo:
Cool and useful addition to the scDVP :)
very small suggestions:
1. Adding the spatial data [ref ](https://www.nature.com/articles/s41592-024-02212-x) in the "4. References" section of the notebook
2. Similarly, adding spatial data integration to "2. Learning Objectives" (maybe something like: Spatial Integration: overlay proteomics data with the original tissue image using SpatialData?)
3. Linking the proteomics table to the cell shapes might be a sensitive step in case the indices do not match. Maybe worth adding a code that verifies/notifies how many shared indices found?

shanibmo (approved):

lucas-diedrich:
Thanks, addressed your comments. I do not `assert` the number of matched observations but I report it now 

```python
matched_elements, matched_table = sd.join_spatialelement_table(
    sdata, spatial_element_names="cells", table_name="proteomics", how="inner"
)
print(f"Found {matched_table.n_obs} shared observations")`
```

PR #242, opened 2026-06-29
add functionality for count-based completeness filtering

vbrennsteiner:
### Purpose of this PR:

Until now, `filter_data_completeness` could only filter by fraction (e.g. 0.5 for tolerating up to half of the values missing), but there was no way to filter for a specific number of missing samples (e.g. allowing for a maximum of 4 samples missing per group). Especially for small groups it is arguably more informative to think of a concrete number of samples missing, as opposed to a certain fraction.

This PR expands the `max_missing` argument: `float` values between 0.0 and 1.0 continue to be interpreted as fractions, but `int` values of 1 and above are interpreted as literal counts. 

### Example:

`filter_data_completeness(adata, max_missing = 0.5)` --> allow for a maximum of 50 % missing values in any group 

`filter_data_completeness(adata, max_missing = 4)` --> allow for a maximum of 4 missing values per group

### Complete cases:

If all features should pass, `max_missing = 1.0` (notice the float) still works as usual. If only fully complete features should pass, `max_missing = 0` and `max_missing = 0.0` both achieve that. 

### Note: 

This change is not expected to break workflows even though the behaviour of `max_missing = 1` (`int`) no longer lets all features pass but filters for maximum one feature missing. The reason is that users are unlikely to run the current float-case of `1.0` is basically a no-op (allowing any degree of missingness is the same as doing nothing).


src/alphapepttools/pp/data.py
```diff
@@ -923,6 +927,35 @@ def _validate_adata_for_completeness_filter(adata: ad.AnnData, action: str, var_
...
+        if max_missing < 0:
+            raise ValueError("Integer threshold must be non-negative.")
+        return True
+    raise ValueError("Threshold must be a float or an int.")
+
+
+def _count_or_fraction_missing(
+    x: np.ndarray,
+    *,
+    count_mode: bool,
```
mschwoer: `is_count_mode`

src/alphapepttools/pp/data.py
```diff
@@ -941,8 +974,8 @@ def filter_data_completeness(
     adata
         AnnData object
     max_missing
-        Maximum fraction of missing values allowed to pass filtering in the interval [0.0, 1.0].
-        Features with a fraction of missing values greater than (`>`) `max_missing` are filtered out.
+        Maximum fraction of missing values allowed to pass. If this is an integer, it is interpreted as the actual number of missing values allowed.
+        If this is a float in the interval [0.0, 1.0], it is interpreted as the fraction of missing values allowed.
```
mschwoer: is this overloading used in one of the canonical libs?
if not, I would go for a boring (mutually exclusive) `max_missing_fraction` & `max_missing_count`

src/alphapepttools/pp/data.py
```diff
@@ -968,6 +1069,18 @@ def filter_data_completeness(
...
+        object, whereas `drop` leaves the input untouched and returns a filtered copy.
+
+    Raises
+    ------
+    TypeError
+        If `adata` is not an AnnData object.
+    ValueError
+        If not exactly one of `max_missing_fraction` / `max_missing_count` is provided, if a
+        threshold is out of range, if `keep_strategy` or `action` is invalid, if `adata` has no
+        features, non-numeric values in `X` or duplicated indices in `obs`, or if a requested
```
mschwoer: (nit) "if a requested
        group is not present in `group_column`." -> `KeyError`?

mschwoer (approved):

PR #243, opened 2026-06-30
resolve all ty diagnostics while keeping invalid-argument-type active

mschwoer:
Add some more type hints.

### DIFF
--- a/src/alphapepttools/_matrix.py
+++ b/src/alphapepttools/_matrix.py
@@ -1,0 +1,31 @@
+"""Internal helpers for accessing AnnData matrix storage.
+
+anndata types ``.X`` and ``.layers[...]`` as a union of every supported storage
+backend (``ndarray | csr_matrix | AwkArray | ZarrArray | CSRDataset | None | ...``).
+The pipeline only ever handles dense numpy arrays, so the union is narrowed in
+one place here rather than at every call site.
+"""
+
+from typing import cast
+
+import anndata as ad
+import numpy as np
+
+
+def get_matrix(adata: ad.AnnData, layer: str | None = None) -> np.ndarray:
+    """Return ``adata.X`` or a named layer as a dense numpy array.
+
+    Parameters
+    ----------
+    adata
+        Annotated data matrix.
+    layer
+        Layer to read. If ``None``, ``adata.X`` is returned.
+
+    Returns
+    -------
+    np.ndarray
+        The requested matrix.
+    """
+    data = adata.X if layer is None else adata.layers[layer]
+    return cast("np.ndarray", data)
--- a/src/alphapepttools/_utils.py
+++ b/src/alphapepttools/_utils.py
@@ -6,10 +6,9 @@ The pipeline only ever handles dense numpy arrays, so the union is narrowed in
 one place here rather than at every call site.
 """
 
-from typing import cast
-
 import anndata as ad
 import numpy as np
+from scipy import sparse
 
 
 def get_matrix(adata: ad.AnnData, layer: str | None = None) -> np.ndarray:
@@ -28,4 +27,6 @@ def get_matrix(adata: ad.AnnData, layer: str | None = None) -> np.ndarray:
         The requested matrix.
     """
     data = adata.X if layer is None else adata.layers[layer]
-    return cast("np.ndarray", data)
+    if sparse.issparse(data):  # this is to cover anndata backends returning sparse data, e.g. csr_matrix
+        data = data.toarray()
+    return np.asarray(data)
--- a/src/alphapepttools/metrics/feature_level.py
+++ b/src/alphapepttools/metrics/feature_level.py
@@ -2,12 +2,14 @@
 
 import numbers
 import warnings
-from collections.abc import Callable, Hashable
+from collections.abc import Callable, Hashable, MutableMapping
 
 import anndata as ad
 import numpy as np
 import pandas as pd
 from scipy.stats import median_abs_deviation
+
+from alphapepttools._matrix import get_matrix
 
 METRICS_KEY = "metrics"
 PMAD_KEY = "pmad"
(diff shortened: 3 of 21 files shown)


src/alphapepttools/pl/plots.py
```diff
@@ -2396,7 +2408,7 @@ def scree_plot(
 
 
 def plot_pca_loadings(
-    data: ad.AnnData | pd.DataFrame,
+    data: ad.AnnData,
```
lucas-diedrich, line 2605: If `pd.DataFrame` is not supported, should we rename the argument to `adata`, despite it being a breaking change?

mschwoer: https://github.com/MannLabs/alphapepttools/pull/265

lucas-diedrich (approved):

src/alphapepttools/_utils.py
lucas-diedrich, line 1: _utils.py to make it more general?

lucas-diedrich (approved):

src/alphapepttools/_matrix.py
```diff
@@ -0,0 +1,31 @@
...
+    layer
+        Layer to read. If ``None``, ``adata.X`` is returned.
+
+    Returns
+    -------
+    np.ndarray
+        The requested matrix.
+    """
+    data = adata.X if layer is None else adata.layers[layer]
+    return cast("np.ndarray", data)
```
lucas-diedrich: I am not sure that it is guaranteed here that the layer is indeed a numpy array - can you make it `np.asarray(data)`?

src/alphapepttools/metrics/feature_level.py
```diff
@@ -146,7 +148,7 @@ def coefficient_of_variation(
 
     """
     adata = adata.copy() if copy else adata
-    data = adata.X if layer is None else adata.layers[layer]
+    data = get_matrix(adata, layer)
```
lucas-diedrich, line 151: nice, thanks for making this a method

src/alphapepttools/metrics/principal_component_regression.py
```diff
@@ -147,7 +149,7 @@ def principal_component_regression(
     """
     _verify_keys__principal_component_regression(adata, covariate=covariate, pca_key=pca_key, pca_uns_key=pca_key_uns)
 
-    pca_embeddings = adata.obsm[pca_key]
+    pca_embeddings = cast("np.ndarray", adata.obsm[pca_key])
```
lucas-diedrich, line 152: np.asarray?

src/alphapepttools/pl/plots.py
```diff
@@ -2703,7 +2713,7 @@ def plot_pca_loadings(
 
 
 def plot_pca_loadings_2d(
-    data: ad.AnnData | pd.DataFrame,
+    data: ad.AnnData,
```
lucas-diedrich, line 2716: argument name: `adata`?

src/alphapepttools/pp/data.py
```diff
@@ -158,7 +160,7 @@ def add_metadata(  # noqa: C901, PLR0912
 
     ### Handle alignment of incoming and existing metadata
     if keep_existing_metadata:
-        existing_metadata = adata.obs if axis == 0 else adata.var
+        existing_metadata = cast("pd.DataFrame", adata.obs if axis == 0 else adata.var)
```
lucas-diedrich, line 163: Just for my understanding, when should one use explicit type hints vs. cast? 

I would have thought that `cast` is mainly necessary, when no variable is set? 

```python
existing_metadata: pd.DataFrame = ... 

# vs. 
...
return cast("pd.DataFrame", ...)
```

lucas-diedrich (approved):
nice, seems like this effort already caught some issues and led to some useful refactors

src/alphapepttools/pl/plots.py
mschwoer: https://github.com/MannLabs/alphapepttools/pull/265

src/alphapepttools/pp/data.py
mschwoer: good question, needed to ask for help on this
```

  The three things do different jobs; only one of them is about types at all.

  1. cast("np.ndarray", x) — a lie to the type checker. Zero runtime effect (it literally return x). Use it when you know something the checker can't prove, and no runtime work is needed. If you're wrong, nothing tells you — the error surfaces 50 lines later as an AttributeError.

  2. x: np.ndarray = ... — a declaration, and it is checked. 

  3. np.asarray(x) — an actual runtime conversion. The return type is genuinely np.ndarray, so the checker is satisfied honestly, and if x is already an ndarray it's a no-op (no copy). 

  Rough decision rule

  - 3.1. Can you fix the annotation upstream (better signature, TypeVar, overload, TypedDict)? Do that. Casts are a symptom of a type you didn't model.
  - 3.2. Is there a cheap runtime conversion that makes the claim true? Use it (np.asarray, list(...), str(...)). Truth beats assertion.
  - 3.3. Do you want the claim enforced? assert isinstance(x, np.ndarray) — narrows for the checker and fails loudly at runtime.
  - 3.4. Only now cast — third-party stubs that are wrong/too wide, Any coming out of an untyped lib, invariance workarounds. Boundary code, ideally one place, with a comment saying why.
```

mschwoer:
this would be the type hints gap: https://github.com/MannLabs/alphapepttools/compare/type-hints_II...type-hints_III?expand=1

lucas-diedrich:
> this would be the type hints gap: https://github.com/MannLabs/alphapepttools/compare/type-hints_II...type-hints_III?expand=1

Ok this would be a bit of an overkill 👍

lucas-diedrich:
@mschwoer is this blocked by something?

PR #244, opened 2026-07-01
Fix/spatialdata plotting: Fix off-by-one error

lucas-diedrich:
Fixed an error in the IO of the shapes that mapped shapes and proteomics data incorrectly to one another.

The spatialdata-io-native function only reads the geometry column of a geojson, leading to a 0-indexed geopandas dataframe. The shape indexing in the Weiss paper is 1-based (see [github repo](https://github.com/BorgwardtLab/Human_scDVP_data_dashboard/blob/eec98a4f9d93a0befaf6483c466d6536e0d11b6a/dashboard/dvp_dashboard.py#L130)). Overall this lead to a mismatch of shapes and measurements which is fixed now by reading the shapes-data (including indices) with geopandas.


lucas-diedrich:
Zonation pattern happens to be better visible 
image

shanibmo (approved):
Good catch :)

PR #246, opened 2026-07-04
Custom limits to MappedColormaps

vbrennsteiner:
### Problem addressed by this PR:

A common task of depicting numerical data is mapping values to colors along a consecutive gradient using a colorbar legend. This presents a key challenge: both the scatterplot and the colorbar need to "know" which data value should be mapped to which color. This is commonly solved with a `matplotlib.ScalarMappable`, which can map a numerical array to a range of colors. 

Our `MappedColormaps` can produce `ScalarMappable` instances with the added benefit of adding a percentile cutoff to the upper and lower ends, providing a solution to outliers dominating the scale and compressing the color mapping of other points (see basic plotting tutorial notebook). 

However, until now, MappedColormaps had a key problem that made it insufficient for producing reusable color mappings: it always mapped the colormap to between the lowest and highest point of the actual data, with no way of setting custom limits. This makes it awkward to produce colorbars, since they start and end at serendipitous values, and it was impossible to fit a colormap to between suitable bounds and easily apply that mapping to data values to be used e.g. in the `color_column` argument of scatter plots, since any call to `fit_transform` would automatically reset the mapping. 

### Fix:

This PR fixes those limitations by 1.) adding a `vmin` and `vmax` argument to `MappedColormaps.fit_transform` to solve the issue with non-ideal ranges, and 2.) adding a `MappedColormaps.transform()` method that only applies an existing scale to data, without re-fitting to the data's scale. 

### Notes:

Existing functionality of MappedColormaps stays untouched, and unit tests cover the new functionality; additionally, a `.fit` method has been added to fit data + custom vmin/vmax values, completing the sklearn-like 'fit/transform/fit_transform' interface.


src/alphapepttools/pl/colors.py
```diff
@@ -253,6 +255,10 @@ def _get_colors_from_cmap(
         colormap and retrieve the corresponding colors in whatever shape the input
         array was. In the case of 2D input arrays, the output will be a mxnx4 array
         of RGBA tuples.
+    vmin : float, optional
+        Minimum value for normalization when values is an array. If None, uses the minimum of the array.
+    vmax : float, optional
```
mschwoer, line 260: I read "maximum velocity here" :-D
`min/max_value`?

src/alphapepttools/pl/colors.py
```diff
@@ -295,7 +301,11 @@ def _get_colors_from_cmap(
     if not pd.api.types.is_numeric_dtype(values):
         raise TypeError("values must be an integer or a numeric numpy array")
 
-    vmin, vmax = np.nanmin(values), np.nanmax(values)
+    if vmin is None:
+        vmin = np.nanmin(values)
```
mschwoer, line 305: (nit)
`vmin = vmin if vmin is not None else np.nanmin(values)`

src/alphapepttools/pl/colors.py
```diff
@@ -902,6 +912,107 @@ def fit_transform(
...
+                d_min, d_max = np.nanmin(arr), np.nanmax(arr)
+        else:
+            d_min = d_max = None
+
+        self.vmin = vmin if vmin is not None else d_min
+        self.vmax = vmax if vmax is not None else d_max
+
+        if self.vmin is None or self.vmax is None:
+            raise ValueError("fit() requires `data`, or both `vmin` and `vmax`, to set the bounds.")
+        return self
```
mschwoer, line 963: do we need to return something?

src/alphapepttools/pl/colors.py
mschwoer, line 792: I think we should move all colormap-related code into a dedicated module

src/alphapepttools/pl/colors.py
```diff
@@ -902,6 +912,107 @@ def fit_transform(
             return np.apply_along_axis(mpl_colors.to_hex, -1, rgba, keep_alpha=True)
```
mschwoer: shouldn't `fit_transform()` just call `fit()` and `transform()` now?

src/alphapepttools/pl/colors.py
vbrennsteiner: Agreed, though I'd put this in a future refactor (see https://mannlabs.atlassian.net/browse/AX-1561?atlOrigin=eyJpIjoiZGNiYjQzODIxNDQxNDcxNjlmYjM1MzdlZWEzYzI5MGIiLCJwIjoiaiJ9)

src/alphapepttools/pl/colors.py
vbrennsteiner: Yes, I wanted to stick with the sklearn template that requires that fit return self, plus it makes the convenient one-liners in study_04 (`mapper = apt.pl.MappedColormaps(cmap="cmc.vik").fit(vmin=lo, vmax=hi)`) and fit_transform (`self.fit(data).transform(data, as_hex=as_hex)`) work

src/alphapepttools/pl/colors.py
```diff
@@ -887,19 +892,105 @@ def fit_transform(
...
+        .. code-block:: python
+
+            mapper = MappedColormaps(cmap="cmc.vik")
+            mapper.fit(vmin=0, vmax=1)
+
+        """
+        if data is not None:
+            arr = np.asarray(data)
+            if self.percentile is not None:
+                d_min = np.nanpercentile(arr, self.percentile[0])
```
mschwoer: `d_min` vs `vmin`  ? :-)

also, could `d` be expanded? `data_min`?

src/alphapepttools/pl/colors.py
```diff
@@ -887,19 +892,105 @@ def fit_transform(
...
+        Bounds are derived from ``data`` (the full range, or the configured
+        ``percentile`` range), unless explicit ``vmin``/``vmax`` are given, which
+        override on a per-bound basis.
+
+        Parameters
+        ----------
+        data : np.ndarray, optional
+            Data from which to derive the bounds. Required unless both ``vmin`` and
+            ``vmax`` are provided. NaNs are ignored.
+        vmin : float, optional
```
mschwoer, line 924: `vmin` -> is this following some convention? if not, could this be `min_value` to be consistent with the code above?

src/alphapepttools/pl/colors.py
```diff
@@ -887,19 +892,105 @@ def fit_transform(
...
+                d_min = np.nanpercentile(arr, self.percentile[0])
+                d_max = np.nanpercentile(arr, self.percentile[1])
+            else:
+                d_min, d_max = np.nanmin(arr), np.nanmax(arr)
         else:
-            self.vmin = np.nanmin(data)
-            self.vmax = np.nanmax(data)
+            d_min = d_max = None
+
+        self.vmin = vmin if vmin is not None else d_min
```
mschwoer: could these be private?

mschwoer (approved):
LGTM, some of the usual naming nits ;-)

src/alphapepttools/pl/colors.py
vbrennsteiner: it's actually matplotlib's convention for mapping colorscales, therefore I thought it best to keep the same syntax

src/alphapepttools/pl/colors.py
vbrennsteiner: hmm not easily, there are applicateions where you'd want to write these deliberately. Also, see the matplotlib comment above..
If they remain public however, I would go ahead and explicitly cover them in the docstring to make the expected behavior clear 👍

PR #248, opened 2026-07-06
Harmonize study 01

vbrennsteiner:
### Purpose of the PR

Until now, the study 01 notebook lacked structured headers and description similar to the study 02 notebook. This PR restructures the  notebook accordingly. Furthermore, principal component regression was so far missing in this study to assess the effects of batch correction and has now been added.


src/alphapepttools/pl/plots.py
```diff
@@ -53,13 +53,19 @@ def _extract_groupwise_plotting_data(
     grouping_column: str | None = None,
     value_column: str | None = None,
     direct_columns: list[str] | None = None,
-) -> tuple[list[list], list[str], list[int]]:
+    subgroup_column: str | None = None,
```
mschwoer, line 57: these changes are from `grouped_abundance_plots` branch, right? could you point the PR that that branch no not have this in the diff?

lucas-diedrich (approved):
LGTM. I would consider again whether we should not support an `axis` argument in this imputation function

vbrennsteiner:
> LGTM. I would consider again whether we should not support an `axis` argument in this imputation function

agreed, that's coming up 👍

PR #251, opened 2026-07-07
Harmonize & expand study 02

vbrennsteiner:
### Purpose of the PR

Standardize headers in study 02 and expand with one example of how `apt.io.mulink_from_anndatas` can be used to generate a linked precursor - protein - genes level `MuData` instance, which in turn can be used to look into how precursors change vs. other data modalities: 

image

Looking at different completely quantified precursor levels, we can see the differential regulation between conditions:

image


shanibmo:
Really cool notebook and nice expansion with the mulink to extract much more from the data :) 

a few comments:
1. I think adding plots to illustrate a consideration or to visualize the analysis done or the could be helpful for non-pros. specifically , adding a histogram of completeness in section 3.1 (to illustrate the fraction of "lost features" when choosing 4/4 complete) and volcano plots (maybe colored by max intensity for each gene) in 3.2-3.4 sections or printing the heads of the DE results (like you did in the mulink) to avoid duplication of the volcano plot after the kinase annotations.
2. maybe worth adding figure legends on the volcanos?
3. for aggregating and computing the 80% you use p-val - is it adjusted p-val? if not, I think it calls for a small explantation in the markdown. 
4. the linkage plot in 4.1.2 is not clear to me - why is it not precursors x proteins plot?

shanibmo (approved):

vbrennsteiner:
> Really cool notebook and nice expansion with the mulink to extract much more from the data :)
> 
> a few comments:
> 
>     1. I think adding plots to illustrate a consideration or to visualize the analysis done or the could be helpful for non-pros. specifically , adding a histogram of completeness in section 3.1 (to illustrate the fraction of "lost features" when choosing 4/4 complete) and volcano plots (maybe colored by max intensity for each gene) in 3.2-3.4 sections or printing the heads of the DE results (like you did in the mulink) to avoid duplication of the volcano plot after the kinase annotations.
> 
>     2. maybe worth adding figure legends on the volcanos?
> 
>     3. for aggregating and computing the 80% you use p-val - is it adjusted p-val? if not, I think it calls for a small explantation in the markdown.
> 
>     4. the linkage plot in 4.1.2 is not clear to me - why is it not precursors x proteins plot?

1. Good points, I think they go in the direction of a future overhaul: once the new linear model based eBayes differential expression is merged, the 100 % completeness filter is actually no longer mandatory - however, the notebook could be expanded to also detail missingness and coverage in more detail. I converted it into a ticket for further notebook enhancement (https://mannlabs.atlassian.net/browse/AX-1516?atlOrigin=eyJpIjoiNzgyNmU2MWFkM2ZjNDg2OThkNWE4ZDQ5ZDkwMWMxZmEiLCJwIjoiaiJ9).

2. On the figure legends - this would be an interesting addition, though we would need to think about other study notebook figures as well. I added a small note in the markdown, though right now none of them have explicit "paper-style" legends, which could be a worthy addition for the study notebook in general (https://mannlabs.atlassian.net/browse/AX-1517?atlOrigin=eyJpIjoiMTk4ZDgzY2NlZTU4NGIwN2EzOGE1MjgwOTY1YzJlMGEiLCJwIjoiaiJ9).

3. Good point on the p-value, I put in a note to the markdown - here it only serves as a rank for scoring the kinases, so the ranking would not change with BH-correction; right now it follows the convention of the original paper which also uses p-value.

4. The linkage plot is all features concatenated: there is one gene, two isoforms, and several precursors concatenated on each axis. The protein-precursor interactions are the second and third columns from the left.

PR #254, opened 2026-07-16
Avoid eager package imports

pjhartout:
## Summary

Importing `alphapepttools` no longer eagerly imports every public submodule, and importing `alphapepttools.pl` no longer loads the heavy plotting implementation until it is accessed. Existing entry points such as `alphapepttools.io`, `alphapepttools.pl.scatter`, and `from alphapepttools import tl` remain available through lazy attribute resolution.

The lazy loaders cache resolved modules and exports after first access, _while dependency errors surface only when the relevant functionality is requested_.
## Dependency boundary

This PR changes import-time behavior without changing the package's declared installation dependencies. Environments that intentionally exclude the analysis dependency stack can still use lightweight plotting primitives, such as `from alphapepttools.pl.figure import AxisManager`, without importing AnnData or Scanpy. Data-aware exports from `alphapepttools.pl.plots` continue to load their analysis dependencies when accessed.

## Implementation note

Lazy module attributes use the module-level `__getattr__` and `__dir__` hooks proposed in [PEP 562](https://peps.python.org/pep-0562/) and documented in the corresponding [Python data model documentation](https://docs.python.org/3/reference/datamodel.html#customizing-module-attribute-access).

This contribution was partially written by an LLM, but I reviewed the changes and this PR description myself.

### DIFF
--- a/src/alphapepttools/pl/__init__.py
+++ b/src/alphapepttools/pl/__init__.py
@@ -1,26 +1,7 @@
+from importlib import import_module as _import_module
+
 from .colors import BaseColormaps, BaseColors, BasePalettes, MappedColormaps, get_color_mapping, show_rgba_color_list
 from .figure import create_figure, label_axes, save_figure
-from .plots import (
-    PlotConfig,
-    Plots,  # Kept here for backward compatibility, using Plots raises and alerts the user about the deprecation and how to upgrade.
-    add_legend_to_axes,
-    add_legend_to_axes_from_patches,
-    add_lines,
-    barplot,
-    boxplot,
-    histogram,
-    label_plot,
-    layered_plot,
-    make_scatter_config,
-    plot_pca,
-    plot_pca_loadings,
-    plot_pca_loadings_2d,
-    rank_median_plot,
-    scatter,
-    scree_plot,
-    violinplot,
-    volcano,
-)
 
 __all__ = [
     "BaseColormaps",
@@ -52,3 +33,21 @@ __all__ = [
     "violinplot",
     "volcano",
 ]
+
+_PLOT_EXPORTS = frozenset(name for name in __all__ if name not in globals())
+
+
+def __getattr__(name: str) -> object:
+    if name == "plots":
+        return _import_module(".plots", __name__)
+
+    if name not in _PLOT_EXPORTS:
+        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+    value = getattr(_import_module(".plots", __name__), name)
+    globals()[name] = value
+    return value
+
+
+def __dir__() -> list[str]:
+    return sorted(set(globals()) | set(__all__) | {"plots"})
--- a/src/alphapepttools/__init__.py
+++ b/src/alphapepttools/__init__.py
@@ -1,5 +1,17 @@
-from . import data, io, metrics, pl, pp, tl
+from importlib import import_module as _import_module
+from types import ModuleType as _ModuleType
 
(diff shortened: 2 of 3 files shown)


lucas-diedrich:
Thanks for your interest in contributing and improving the package @pjhartout 

- Could you please share a bit more about the pain points that you had when using alphapepttools? When did you in particular face issues with the dependency resolution/installed packages?

> This contribution was partially written by an LLM, but I reviewed the changes and this PR description myself.

Could you please share the prompt that generated the changes?

- I haven't seen  the utilized lazy import pattern yet - could you point me to an example/a reference implementation that implements this pattern and motivated the design?

pjhartout:
> Could you please share a bit more about the pain points that you had when using alphapepttools? When did you in particular face issues with the dependency resolution/installed packages?

I have a rather contrived use case where I am only interested in some of the primitives of the package such as `AxisManager` that don't depend on `scanpy` or `AnnData`. In my project, I will do the equivalent of `pip install --no-deps alphapepttools` and just import `AxisManager` with `from alphapepttools.pl.figure import AxisManager` (I also want to use other classes, this is just an example). Like I say in the PR description, the changes proposed here allow me to do this without having to import libraries that are not relevant for this functionality.

> Could you please share the prompt that generated the changes?

I basically described the answer to your first question above (i.e. my goal of just importing the functionality I need without the dependencies), and this is what it came up with. As per my personal instructions, the LLM then surfaced some choices and options, and I chose from them. There was not a single prompt but a multi-turn conversation over several sessions which is rather hard to convey here. I also switched harnesses midway, which might affect results too. I'm happy to share more in another setting, though. Overall I'm quite happy with those changes, since they are quite minimal while still allowing me to use `alphapepttools` the way I need it to work, i.e. without duplicating the classes I need and without having to import all the dependencies the package comes with.

> I haven't seen the utilized lazy import pattern yet - could you point me to an example/a reference implementation that implements this pattern and motivated the design?

Several relevant sources here, and this is a rather pythonic pattern:
- [SPEC 1](https://scientific-python.org/specs/spec-0001/) 
- [scipy](https://github.com/scipy/scipy/blob/20a186fb9611ff4787f0fd6d8784d42339fdcd27/scipy/__init__.py#L95-L141)
- [pydantic](https://github.com/pydantic/pydantic/blob/92462ee5a5200e8f21825023b83fccb9d4ee63c5/pydantic/__init__.py#L248-L456)
- [scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/8816591844d858f8a2dcb212e47be13ddd1ac8e1/sklearn/__init__.py#L73-L136)
They seem to motivate it partially for import speeds, but I don't know if that's relevant for you. I guess your import times improve indirectly as a result.

pjhartout:
@lucas-diedrich I fixed some of the copilot reviews and the coverage issues that were raised.

src/alphapepttools/pl/__init__.py
```diff
@@ -1,26 +1,8 @@
+from importlib import import_module
+from types import ModuleType
+
 from .colors import BaseColormaps, BaseColors, BasePalettes, MappedColormaps, get_color_mapping, show_rgba_color_list
 from .figure import create_figure, label_axes, save_figure
-from .plots import (
```
mschwoer, line 3: does the reason for restricting the view on the `plots` submodule like this still hold @vbrennsteiner  ? if so, we should reinstall it

pjhartout: I can revert this, I just did it to avoid drift in the future but if it's a mature set of utils there's no need to worry about this.

PR #267, opened 2026-08-06
[fix] Adapt to new mudata convention

lucas-diedrich:
Mudata change~d~s the settings override context manager from a module to a method in version 0.4
see: https://mudata.readthedocs.io/latest/changelog.html

Discussed in #239


vbrennsteiner:
> Mudata changed the settings override context manager from a module to a method in version 0.4 see: https://mudata.readthedocs.io/latest/changelog.html
> 
> Discussed in #239

:smiling_face_with_tear:

vbrennsteiner (approved):
LGTM!

We could put a TODO here to eventually drop the backwards compatibility when the main MuData version has moved up?

PR #268, opened 2026-08-20
[fix] Fix return type of function

lucas-diedrich:
ty complains about the return type of the function


vbrennsteiner (approved):
LGTM, came up for me a couple times as well

PR #269, opened 2026-08-20
[Fix] Update principal component regression key defaults

lucas-diedrich:
Mentioned by @Matthias-MSMS


vbrennsteiner (approved):
LGTM, good to have the defaults match 👍

PR #270, opened 2026-08-21
[Doc] API consistency and `copy` vs. `inplace` usage

lucas-diedrich:
Document API behavior for reused keywords `copy` vs. `inplace`. This PR just proposes to formalize these conventions in `Contributing.md`

## Suggestion
- `copy` is used when the function acts on the measurement data (`.X`, `.layers`) ("The function either returns a modified anndata object or a new copy of an anndata object")
- `inplace` is used when the function generates summary statistics ("The function either modifies the anndata object or returns a dataframe")

## Implications
The current API is inconsistent in the following functions (review by Claude):

### Return types
- [x] `metrics.pooled_median_absolute_deviation` always returns `pd.DataFrame`, independent of `inplace`
- [ ] `metrics.total_intensity`, `metrics.number_detected`, `metrics.fraction_complete` return `np.ndarray` instead of `pd.DataFrame`

### Differential expression
Comment: This is known and on the roadmap, left it here to inform the refactor. @vbrennsteiner 
- [x] `tl.diff_exp_ebayes` returns `tuple[str, pd.DataFrame]`, `tl.diff_exp_alphaquant` returns `tuple[str, dict[str, pd.DataFrame]]`
- [ ] `tl.diff_exp_ttest` is annotated `pd.DataFrame | None`, but `None` is unreachable (validation raises) and collides with the "modified inplace" meaning of `None`
- [ ] `tl.diff_exp_ttest`, `tl.diff_exp_ebayes`, `tl.diff_exp_alphaquant` store nothing, no `inplace` and no `layer`

### Wrong/inconsistent keyword
- [ ] `metrics.coefficient_of_variation(..., copy=)` should be `inplace` **Breaking**
- [ ] `metrics.principal_component_regression` returns a `float` only, no `inplace`, result cannot be stored
- [ ] `layer` is positional in `pp.nanlog`, `pp.normalize`, `pp.scale_and_center`, `pp.scanpy_pycombat`, `pp.impute_gaussian`, `pp.impute_median`, `pp.impute_knn`, `tl.pca`, `tl.bpca` — must be keyword-only **Breaking**

### Missing `copy`
- [x] `pp.add_metadata` always returns a new object
- [ ] `pp.filter_by_metadata` always returns
- [ ] `pp.filter_data_completeness` mutates and returns the *same* object for `action="flag"`, but returns a copy for `action="drop"`
- [ ] `pp.drop_singleton_batches` always copies (and copies twice internally)
- [ ] Shape-changing functions (`pp.filter_*`, `pp.add_metadata`, `pp.drop_singleton_batches`) do not have a copy argument right now

### Inconsistent result key
- [ ] `key_added` (`metrics.coefficient_of_variation`), `column` (sample-level metrics), `var_colname` (`pp.filter_data_completeness`), `embeddings_name` (`tl.pca`/`tl.bpca`), `pca_key`/`pca_key_uns` (`metrics.principal_component_regression`) — standardize on `key_added`
- [ ] `metrics.calculate_qc_metrics` overrides every default column name of the functions it calls (`total_intensity` → `total_sample_intensity`, `number_detected` → `num_features_detected`, `fraction_complete` → `fraction_detected_features`)

### Bugs
- [ ] `pp.scanpy_pycombat(..., copy=False)` silently does nothing if the batch column contains NaNs: `coerce_nans_to_batch` rebinds `adata` to an internal copy, so the correction is discarded and `None` is returned
- [ ] `pp.filter_by_metadata` returns a *view*; a subsequent inplace `pp` call writes through into the parent object (anndata `FutureWarning`: will raise in future versions)

### Open decisions
- [ ] `tl.pca`/`tl.bpca` write only `.obsm`/`.varm`/`.uns`, so the new taxonomy puts them in the summary-statistics category (`inplace`), but both expose `copy` for scanpy compatibility — needs an explicit carve-out in the guide **Breaking**
- [ ] `pp.detect_special_values` is public in `pp` but takes `np.ndarray`, not `AnnData` — make it AnnData-facing or private

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -42,15 +42,18 @@ pip install -e ".[dev,test,doc]"
 
 ## Handling anndata objects
 
-The central data structure of `alphapepttools` is the `anndata.AnnData` object. All functions should be compatible with `anndata.AnnData`.
+The central data structure of `alphapepttools` is the `anndata.AnnData` object. All functions should be compatible with `anndata.AnnData`. We default to `inplace` modifications, i.e. functions should default to returning `None` and acting on passed `anndata.AnnData` objects directly. This behaviour is adapted from [`scanpy v1`](https://scanpy.readthedocs.io/en/stable/) and aims to maximize the compatibility of the interfaces.
 
-Functions that act on the omics data in the anndata object (typically in the `.pp` and `.tl` modules) should generally follow the following call signature
+### Acting on omics measurements
+
+Functions that act on the measurement data (`.X`, `.layers[...]`) in the anndata object (typically in the `.pp` and `.tl` modules) or change the object's shape (e.g. filtering operations) _MUST_ follow the following call signature
 
 ```python
-alphapepttools.pp.func(adata: ad.AnnData, ..., layer: str | None = None, copy: bool = False) -> None | ad.AnnData:
+alphapepttools.pp.func(adata: ad.AnnData, ..., *, layer: str | None = None, copy: bool = False) -> None | ad.AnnData:
 ...
 
-alphapepttools.tl.func(adata: ad.AnnData, ..., layer: str | None = None, copy: bool = False) -> None | ad.AnnData:
+alphapepttools.tl.func(adata: ad.AnnData, ..., *, layer: str | None = None, copy: bool = False) -> None | ad.AnnData:
+  ...
 
 ```
 
@@ -58,9 +61,7 @@ alphapepttools.tl.func(adata: ad.AnnData, ..., layer: str | None = None, copy: b
 
 **Modification inplace** Per default, the `anndata.AnnData` object is modified inplace (`copy=False`), this means that the current object is updated and the function returns `None`. If `copy=True`, an updated copy of the object is returned and the original object remains unchanged.
 
-This behaviour is adapted from [`scanpy`](https://scanpy.readthedocs.io/en/stable/) and aims to maximize the compatibility of the interfaces.
-
-### Examples
+#### Examples
 
 Default behaviour:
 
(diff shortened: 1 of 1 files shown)


vbrennsteiner (approved):
Valiant effort, this is really cool! These inconsistencies have been flaring up here an there in the past but so far we never had them all addressed in one cohesive review. I agree with the contributing guidelines, but what should we do about the existing issues? The PR description is highly valuable at this point, should we convert these tasks to tickets before merging?

docs/contributing.md
```diff
@@ -98,7 +99,30 @@ assert not np.array_equal(adata.X, adata_new.X)
...
+
+```python
+alphapepttools.tl.func(adata: ad.AnnData, ..., *, layer: str | None = None, inplace: bool = True) -> None | pd.DataFrame:
+  ...
+
+alphapepttools.metrics.func(adata: ad.AnnData, ..., *, layer: str | None = None, inplace: bool = True) -> None | pd.DataFrame:
+  ...
+```
+
+In summary, the keyword `inplace` also determines if an anndata object is modified, but it determines if a result is added to a non-measurement layer slot in anndata (`.obs`, `.var`, `.obsm`, `.varm`, `.uns`) or if it is returned as dataframe.
```
mschwoer: did not get this sentence :-)

docs/contributing.md
```diff
@@ -98,7 +99,30 @@ assert not np.array_equal(adata.X, adata_new.X)
...
+
+#### Examples
+
+```
+result = apt.tl.func(adata, ..., inplace=False)
+assert isinstance(result, pd.DataFrame)
 
+result = apt.tl.func(adata, ..., inplace=True)
+assert result is None
```
mschwoer, line 133: would it hurt (us and/or the convention) if we always return the `df` ?

docs/contributing.md
lucas-diedrich: I got a stroke reading this again

docs/contributing.md
lucas-diedrich: I think it's unintuitive to get a result _and_ have an inplace change of the anndata object as side effect

mschwoer: as discussed in person: I am find with this

lucas-diedrich:
@vbrennsteiner Added the tasks as tickets, would just go work through them one by one. I would wait with merging this PR until the most serious issues got resolved.

docs/contributing.md
lucas-diedrich: fixed

lucas-diedrich:
@vbrennsteiner As discussed in person, we should streamline everything to a single argument (`inplace`).

## Suggestion 
Suggested migration. 

### Next minor version (version +0.1.0)
For every function that uses `copy`: 

1. Add `inplace` as additional argument with the same default behavior as `copy` (equivalent to `inplace = (not copy)`. 
2. Extend supported type for copy to `bool | None`. Update default for `copy` to `None` (previously True/False boolean). 
3. If copy is explicitly set (equivalent to `if copy is not None`): Copy takes precedence over `inplace`. Warn users that copy argument is deprecated and will be remvoed in next minor version in favor of `inplace`. Internally map copy behavior to `inplace`. 

_Optional_ I would further suggest to streamline the `inplace` behavior across all functions in the package with `inplace=True` and warn about this 

### In 2 minor versions (version +0.2.0)
1. Remove copy argument. 

_Optional_ Should we add a custom message for the copy argument ("got removed in...") or just raise the native python TypeError? 


### Suggested implementation
Function decorator?

PR #271, opened 2026-08-22
[Docs] Tutorial 2: Minimal changes to use mulink-internal plotting

lucas-diedrich:
Builds on #251. Drop 5 lines of code by using `link.pl.adjacency_matrix` accessor instead of custom plots


vbrennsteiner (approved):
LGTM!

PR #273, opened 2026-09-01
harmonizing differential expression output

vbrennsteiner:
### PR for harmonizing differential expression output

As discussed with @lucas-diedrich, differential expression results should be standardized to allow integrating the new empirical Bayes moderated t-test implementation, as well as standardized downstream visuals that consume differential expression outputs.

Right now, differential expression functions have diverging output formats. This is a bad starting point for implementing new differential expression modules, and a problem for downstream visualizations depending on standardized results.

- ttest returns the dataframe directly `result_df`
- AlphaQuant returns the tuple (`comparison_name`, `results_dict`), where the dict contains "protein" "proteoform" and "peptide" result dataframes. 
- Current EBayes returns the tuple (`comparison_name`, `results_df`)

**The goal of this PR is to apply minimal changes to harmonize all three methods to return a single dataframe with standardized columns + method specific columns** (like e.g. proteoform values for AlphaQuant)


src/alphapepttools/tl/diff_exp/alphaquant_wrapper.py
```diff
@@ -133,11 +178,20 @@ def diff_exp_alphaquant(
...
     -------
-    tuple[str, dict[str, pd.DataFrame]]
-        Tuple containing:
-        - comparison_key: String identifier for the comparison (e.g., "Group1_VS_Group2")
-        - results: Dictionary with keys 'protein', 'peptide', 'proteoform' containing
-          standardized differential expression results for each level.
+    pd.DataFrame
+        Standardized results for all three analysis levels stacked into a single frame.
+        The `modality` column holds 'protein', 'proteoform' or 'peptide' and is used to
+        select the level of interest; the `feature` column holds each row's own
```
lucas-diedrich: Rename to `feature_id` column?

src/alphapepttools/tl/diff_exp/alphaquant_wrapper.py
```diff
@@ -103,6 +118,36 @@ def _standardize_alphaquant_results(
...
+    """
+    stacked = []
+    for level, df in results.items():
+        level_df = df.copy()
+        level_df["modality"] = level
+        level_df["feature"] = level_df[_FEATURE_COLUMN_PER_LEVEL[level]]
+        stacked.append(level_df)
+
+    # reindex both enforces the column order and pads the level-specific columns with NaN
+    return pd.concat(stacked, axis=0, ignore_index=True).reindex(columns=_STACKED_COLS)
```
lucas-diedrich, line 149: Equivalent:
```python
(
        pd.concat(
            {level: df.set_index(_FEATURE_COLUMN_PER_LEVEL[level], drop=False) 
             for level, df in results.items()}, axis=0
        )
        .reset_index(names=["modality", "feature"])
        .reindex(columns=_STACKED_COLS)
    )
```

Why do you keep the original column? Wouldn't it be better to only keep the unified name (`feature`/`feature_id`)?

tests/integration/test_alphaquant_pipeline.py
```diff
@@ -53,65 +53,31 @@ def test_diff_exp_alphaquant__integration(test_data: tuple[ad.AnnData, pd.DataFr
...
+        "log2fc",
+        "p_value",
+        "-log10(p_value)",
+        "fdr",
+        "-log10(fdr)",
+        "method",
+        "max_level_1_samples",
+        "max_level_2_samples",
+        "quality_score",
+        "proteoform_id",
```
lucas-diedrich, line 72: `proteoform_id` and `sequence` are redundant with the `feature` column for the respective subsetted data frames, you could consider to remove them from the `_STACKED_COLS` variable, but I guess that's fine.

lucas-diedrich (approved):
thanks for standardizing!

src/alphapepttools/tl/diff_exp/alphaquant_wrapper.py
vbrennsteiner: I kept the originals in the interest of keeping AlphaQuant's original output around for users who might be migrating their alphaquant workflows and looking for the familiar column here. This way, filtering for the correct modality directly gives back recognisable AlphaQuant outputs. Initially, I wasn't even sure if a collective feature column would make sense but ultimately having a feature column without any nans seemed worth it..

For the stacking, I'd favor the current over the multi-index solution in this case. While the new solution does everything in one command, building the multiindex and getting rid of it again takes more time, has about as many (linted) lines as the current one, while being harder to read.

tests/integration/test_alphaquant_pipeline.py
vbrennsteiner: true - let's keep them around just for now and see how users interact with that output, I added a corresponding TODO to log the decision

PR #274, opened 2026-09-01
[feature] Allow for center/scale only in `apt.pp.scale_and_center` 

lucas-diedrich:
Extend functionality of scaling preprocessing step, as used in the multimodal notebook. Especially in TMT experiments, users often want to median-center (but not scale). This is now one special case of the function

This PR
1. Updates the function and allows users to optionally  center-only or scale-only.
```
apt.pp.scale_and_center(..., center=True/False, scale=True/False)
```
3. Adds tests
4. Use instead of custom median-centering logic in multimodal notebook (validated that the arrays are exactly numerically equivalent)

### DIFF
--- a/src/alphapepttools/pp/data.py
+++ b/src/alphapepttools/pp/data.py
@@ -810,22 +810,34 @@ def data_columns_to_df(
 
 
 def scale_and_center(  # explicitly tested via test_pp_scale_and_center()
-    adata: ad.AnnData, scaler: str = "standard", layer: str | None = None, *, copy: bool = False
+    adata: ad.AnnData,
+    scaler: Literal["standard", "robust"] = "standard",
+    layer: str | None = None,
+    *,
+    center: bool = True,
+    scale: bool = True,
+    copy: bool = False,
 ) -> None | ad.AnnData:
-    """Scale and center data.
-
-    Either use standard or robust scaling. 'robust' scaling relies
-    on interquartile range and is more resistant to outliers. Scaling
-    operates on columns only for now.
+    """Scale and center features
 
     Parameters
     ----------
     adata
         AnnData object with data to scale.
     scaler
-        Sklearn scaler to use. Available scalers are 'standard' and 'robust'.
+        Sklearn scaler to use. Available scalers are
+            - `standard`: Mean centering and scaling by standard deviation
+            - `robust`: Median centering and scaling by provided quantiles.
     layer
         Name of the layer to scale. If None (default), the data matrix X is used.
+    center
+        Whether to center the data distribution around zero before scaling. If True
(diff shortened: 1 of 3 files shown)


src/alphapepttools/pp/data.py
```diff
@@ -854,20 +869,39 @@ def scale_and_center(  # explicitly tested via test_pp_scale_and_center()
...
+
+    See Also
+    --------
+    :class:`sklearn.preprocessing.StandardScaler`
+    :class:`sklearn.preprocessing.RobustScaler`
 
     """
+    if not (center or scale):
+        raise ValueError(
+            "Setting `center=False` and `scale=False` corresponds to an identity transform. Set at least one argument to `True`"
```
vbrennsteiner: corresponds to an identity transform is technically correct, but it could also be said explicitly as "does nothing" or maybe "leaves data unchanged"? 
Also one could consider whether this raise is even necessary, considering that these are keyword arguments that have to be deliberately set, and since it's not a "failure" per se e.g. if it's part of a pipeline that decides on the fly whether something should be centered and/or scaled or none of the two. Then again, users could simply gate calling the function in the first place, so I guess it comes down to taste

src/alphapepttools/pp/data.py
```diff
@@ -854,20 +869,39 @@ def scale_and_center(  # explicitly tested via test_pp_scale_and_center()
...
+        )
     adata = adata.copy() if copy else adata
     logging.info(f"pp.scale_and_center(): Scaling data with {scaler} scaler.")
 
     if scaler == "standard":
-        scaler = StandardScaler(with_mean=True, with_std=True)
+        scaler = StandardScaler(with_mean=center, with_std=scale)
     elif scaler == "robust":
-        scaler = RobustScaler(with_centering=True, with_scaling=True, quantile_range=(25.0, 75.0))
+        scaler = RobustScaler(with_centering=center, with_scaling=scale, quantile_range=(25.0, 75.0))
```
vbrennsteiner: Are these values documented somewhere else, or could they maybe be defined upfront for transparency?

vbrennsteiner (approved):
Nice addition, LGTM!

src/alphapepttools/pp/data.py
lucas-diedrich: Much better phrasing! I raised an error as it is unexpected that a function does nothing

src/alphapepttools/pp/data.py
lucas-diedrich: Described in the docstring at the `scaler` and `scale` arguments 

I considered exposing the argument, but then I thought that this adds unnecessary complexity as only certain scalers have extra args.

PR #275, opened 2026-09-03
[API] Fix PMAD return signature

lucas-diedrich:
PMAD + PCV returned dataframe independent of `inplace` parameter


vbrennsteiner (approved):
LGTM!

PR #279, opened 2026-09-05
[fix] Fix key issue in tutorial 00

lucas-diedrich:
The tutorial used a non-existing protein key for the mulink demo, which led to failing tests. Not sure why this only popped up now?


vbrennsteiner (approved):
hmm weird, I thought it was already P01011 but apparently not - thanks for finding/fixing! LGTM!

PR #281, opened 2026-09-08
[refactor] Speedup anndata factory

lucas-diedrich:
The current anndata factory has to be initialized every time a new feature level is requested because feature level-specific columns `feature_id_column` and `intensity_column` are passed to the class at initialization time. 

For large PSM reports, the I/O is expensive and can take multiple minutes. This PR refactors the anndata factory so that it is initialized _only_ with a standardized alphabase report and feature-level specific columns are only required at pivoting time. 

## Results
For the Albrecht PCA-N report, this reduces overhead significantly if 3 feature-levels are requested:
- Before: 3x 3 min I/O ~ 9min for the creation of precursor, protein, gene level anndata objects
- Now: 1x 3 min I/O + (6s + 6s + 6s) pivoting operations ~ 3:20 min for creation of all 3 anndata objects (~3-fold speed up)


vbrennsteiner (approved):
Thanks for fixing this, that's been a long time coming :) 
LGTM!

